PR #30, opened 2015-11-30
stats via HTTP/HTTPS POST

hbellur:
Based on configuration the changes will allow user stats to be sent to a
server, after a session has ended.

The following information is posted

username
received packets
received bytes
sent packets
sent bytes

The changes have been tested against OpenSSL 1.0.2a

### DIFF
--- a/examples/etc/turnserver.conf
+++ b/examples/etc/turnserver.conf
@@ -620,3 +620,22 @@
 #no-tlsv1
 #no-tlsv1_1
 #no-tlsv1_2
+#
+# Send stats via HTTP/HTTPS POST. By default, no stats are sent.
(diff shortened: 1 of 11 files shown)


mom040267:
Unfortunately, I cannot use that merge. The change uses synchronous network operations (connect, write). This is a multi-session server, it is handling many sessions in a few threads, so all operations must be asynchronous. See how it is done in other TCP communications in the TURN server. Please fix it.

mom040267:
The proposed changes are violating our communications pattern - all TCP
communications must be asynchronous - because many sessions are handled in
every thread. I asked to fix that before we can use it.

Thanks
Oleg

PR #60, opened 2016-08-26
Set "READ_TIMEOUT" just after coturn sets "CONNECT_TIMEOUT".

oolong-high:
For MYSQL_OPT_READ_TIMEOUT of variable does not exist, it has been substituting the "connect_timeout".

[Assumed cause]
Because "MYSQL_OPT_READ_TIMEOUT" is not set, coturn cannot detect mysql_ping timeout.
I think coturn shoud have "READ_TIMEOUT" setting(like tomcat) to detect mysql_ping timeout.

[Possible modification]
Set "READ_TIMEOUT" just after coturn sets "CONNECT_TIMEOUT".


misi:
I think it would be better to create a new separated variable for read_timeout and not reuse connect_timeout.
Could you please do it this way?

mom040267:
Thanks for the idea. This work actually includes more changes than that. I committed the changes to the master branch.

Issue #63, opened 2016-09-16
Improve configurability


akatsukle:
We're running TURN tests using the coturn.
For unit testing purposes, it would be great if it was possible to configure the different timers inside coturn.
For example, using the flag `--stale-nonce` allows us to change the nonce every 10 minutes, as defined in 
ns_turn_msg_defs.h at line 68. I would like to pass an additionnal parameter to set the value here.

It would be useful for unit tests, but would also be valid considering the RFC for the following : 
- STUN_MAX_ALLOCATE_LIFETIME (3600)
- STUN_NONCE_EXPIRATION_TIME (600)

For unit testing purposes, it would be nice to extend it also to : 
- STUN_DEFAULT_ALLOCATE_LIFETIME (600)
- STUN_CHANNEL_LIFETIME (600)
- STUN_PERMISSION_LIFETIME (300)

I can make a pull request for it if you greenlight it.

mom040267:
Please make a pull request, and we will review it.

Thanks
Oleg

misi:
Many Thanks for the Pull Requests. 
For me it seems to be OK, but let's wait for Oleg review..

mom040267:
I ll check it this week

Thanks
Oleg

mom040267:
these changes require documentation changes. I'll work on it.

thanks
Oleg

mom040267:
I merged everything, please double-check that everything is correct.

Thanks
Oleg

akatsukle:
The merge looked like correct, thanks.
However, I missed a bit on parameters initialization. I opened a pull request with a patch for it. (#70)

PR #65, opened 2016-09-20
Improve configurability: max allocation lifetime

akatsukle:
This allow to set the duration before the nonce will be invalidated and send a 438 error message.
This is part of #63, as to be able to perform unit testing. It is also RFC compliant since it is only recommended that the maximum allocate lifetime does not exceed one hour. With this setting it is also possible to reduce the maximum allocate lifetime.
The first commit that is necessary for all branches should not conflict upon merging as they are all based on the same commit.

In many functions, I had to add a parameter `max_lifetime` to specify the max allocated lifetime. This was necessary because the check was done inside the function that construct the packet instead of the caller. I did not wanted to change the implementation so I went with it, but it is probably a good idea to move it elsewhere in the future.

src/apps/relay/mainrelay.c:127 will conflict upon merging several pull requests, because of the initialisation of the parameters. Upon each conflict, a zero should be added on this line.


mom040267:
I merged the changes into master

PR #66, opened 2016-09-20
Improve configurability: channel lifetime

akatsukle:
This allow to set the duration of a permission upon the reception of a channel-bind request.
This is part of #63, as to be able to perform unit testing. It is not RFC compliant since this value should always be 600 secs (10 min). This modification has for only purpose to perform testing and debugging.
The first commit that is necessary for all branches should not conflict upon merging as they are all based on the same commit.

src/apps/relay/mainrelay.c:127 will conflict upon merging several pull requests, because of the initialisation of the parameters. Upon each conflict, a zero should be added on this line.


mom040267:
I merged it into master

PR #70, opened 2016-09-26
fix #63 issue

### LINKED ISSUE #63: Improve configurability

akatsukle:
Fix an issue with parameter initialization in the previous pull requests.
Previously it was working fine as long as the parameter was passed on the command line, but would fallback on 0 instead of the default value when the parameter was absent.

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -124,7 +124,24 @@ LOW_DEFAULT_PORTS_BOUNDARY,HIGH_DEFAULT_PORTS_BOUNDARY,0,0,0,"",
 /////////////// stop server ////////////////
 0,
 /////////////// MISC PARAMS ////////////////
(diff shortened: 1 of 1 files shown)


mom040267:
I merged that, thanks !

Oleg

PR #73, opened 2016-10-03
fix memory initialization in src/client/ns_turn_msg.c in function stun_attr_get_addr_str

akatsukle:
The initialization was on the memory we received in parameter, but this memory will get copied later in map_addr_from_public_to_private.
However the source of the copy, public_addr is never initialized.

### DIFF
--- a/src/client/ns_turn_msg.c
+++ b/src/client/ns_turn_msg.c
@@ -1439,7 +1439,8 @@ int stun_attr_get_addr_str(const u08bits *buf, size_t len, stun_attr_ref attr, i
   stun_tid_from_message_str(buf, len, &tid);
   ioa_addr public_addr;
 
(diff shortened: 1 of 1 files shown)


mom040267:
No, that is not exactly what is happening:

1) we do have to initialize ca because the function may exit before copying public_addr to ca, and the calling code may rely on it.
2) public_addr will be set - in stun_addr_decode(...).

I guess both ca and public_addr must be initialized.

akatsukle:
Indeed you are right, the function might return before the copy so we have to initialize both.

akatsukle:
I updated the commit to squash them into one.
You can review the pull request.

mom040267:
I'll merge the new version but then I'll change ns_bzero(...) to addr_set_any(...)

PR #96, opened 2017-01-19
Function returns so variable to exit the while loop is never read

chadbrewbaker:
Cleanup on a dead store.


mom040267:
I fixed this problem in a different way. Thanks !
Oleg

PR #114, opened 2017-03-23
Typo in the ALPN callback: ta_len should be ha_len

### LINKED ISSUE #113: socket buffer operation error when using TCP/TLS
While setting up a `turns:` server I noticed that I could not connect to it using Google's webrtc library [1].

> 2721: IPv4. tcp or tls connected to: 1.2.3.4:47716
> 2721: session 000000000000000002: TLS/TCP socket disconnected: 1.2.3.4:47716
> 2721: session 000000000000000002: closed (2nd stage), user <> realm <mydomain> origin <>, local 192.168.43.70:8443, remote 1.2.3.4:47716, reason: **TLS/TCP socket buffer operation error** (callback)

In order to guess what the problem is I tried to connect to coturn via https using Chromium, but the browser gave me this error: `ERR_SSL_PROTOCOL_ERROR` , while coturn logged the same error as before. After trying a few things I noticed that I could connect to the server using Firefox, curl, or other tools, but not with Chromium. If I don't use TLS everything works fine both with Chromium and with the webrtc library.

I really don't know what could be wrong. The only thing I can think of if that both Chromium and the webrtc library use Google's BoringSSL library instead of OpenSSL, so the culprit may lie there,

This issue is easy to reproduce, for example this is a coturn server working fine with curl
```
$ curl https://turn1.whispersystems.org
<!DOCTYPE html>
<html>
  <head>
    <title>TURN Server (https admin connection)</title>
 <style> table, th, td { border: 1px solid black; border-collapse: collapse; text-align: left; padding: 5px;} table#msg th { color: red; background-color: white; } </style> </head>
  <body>
    <b>TURN Server</b><br><i>https admin connection</i><br>
<br>To use the HTTPS admin connection, you have to set the database table <b><i>admin_user</i></b> with the admin user accounts.<br>
</body>
</html>
```
the same is true with Firefox, while Chromium gives `ERR_SSL_PROTOCOL_ERROR`. I'm not the only one experiencing this issue [2] [3].

I'm running coturn 4.5.0.5 from Debian Stretch.

[1] https://chromium.googlesource.com/external/webrtc
[2] https://groups.google.com/forum/#!topic/turn-server-project-rfc5766-turn-server/CxrWCQjDGKc
[3] https://groups.google.com/forum/#!topic/turn-server-project-rfc5766-turn-server/-jFdNKSdebE

paride:
Fixes #113.
Credit for the fix goes to @davidben.

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -2456,7 +2456,7 @@ static int ServerALPNCallback(SSL *ssl,
 		}
 		if((current_len == ha_len) && (memcmp(ptr+1,HTTP_ALPN,ha_len)==0)) {
 			*out = ptr+1;
-			*outlen = ta_len;
+			*outlen = ha_len;
 			SSL_set_app_data(ssl,HTTP_ALPN);
 			found_http = 1;
 		}


misi:
Great catch! Thx!

mom040267:
Thanks guys !

Oleg

PR #160, opened 2017-07-03
config warning about both using use-auth-secret and lt-cred-mech

skystar-p:
I found that if I use `lt-cred-mech` and `use-auth-secret` option both, only `use-auth-secret`'s behavior is appeared and credentials in `turnusers_lt` table are disabled.

I think this should be alarmed with warning.

Thanks.


misi:
Many thanks for the PR

PR #176, opened 2017-09-16
Add --prod flag to EXTRA_OPTIONS in sysconfig file

LorbusChris:
Defaults should be as secure as possible

### DIFF
--- a/rpm/turnserver.sysconfig
+++ b/rpm/turnserver.sysconfig
@@ -2,4 +2,4 @@
 # TURN Server startup options
 #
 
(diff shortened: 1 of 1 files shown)


misi:
Thanks for your contribution, but AFAIU to make default to hide version is not appropriate.
coTURN works standards compliant, it's actual behaviour is correct according the current RFC.

See: https://tools.ietf.org/html/rfc5389#section-15.10
 Its value SHOULD include manufacturer and version number.  

If you agree, then please close this PR

PR #178, opened 2017-09-29
NAT Discovery improvements

misi:


misi:
I will refactor it to avoid conflict

PR #191, opened 2017-11-20
Report total traffic used per allocation via Redis

bradleythughes:
Publish total traffic information when an allocation is deleted. This
makes it easier for applications to monitor total usage per session.

### DIFF
--- a/src/apps/relay/ns_ioalib_engine_impl.c
+++ b/src/apps/relay/ns_ioalib_engine_impl.c
@@ -3596,6 +3596,14 @@ void turn_report_allocation_delete(void *a)
 					}
 					send_message_to_redis(e->rch, "del", key, "");
 					send_message_to_redis(e->rch, "publish", key, "deleted");
+
+					// report total traffic usage for this allocation
+					if(ss->realm_options.name[0]) {
+						snprintf(key, sizeof(key), "turn/realm/%s/user/%s/allocation/%018llu/total_traffic", ss->realm_options.name, (char*)ss->username, (unsigned long long)ss->id);
+					} else {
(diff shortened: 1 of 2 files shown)


misi:
Thanks for PR. 
I have checked it, and I am Ok with it, but let's wait @mom040267 review.

bradleythughes:
Sounds good :)

src/apps/relay/ns_ioalib_engine_impl.c
```diff
@@ -3596,6 +3596,14 @@ void turn_report_allocation_delete(void *a)
 					}
 					send_message_to_redis(e->rch, "del", key, "");
 					send_message_to_redis(e->rch, "publish", key, "deleted");
+
+					// report total traffic usage for this allocation
+					if(ss->realm_options.name[0]) {
+						snprintf(key, sizeof(key), "turn/realm/%s/user/%s/allocation/%018llu/total_traffic", ss->realm_options.name, (char*)ss->username, (unsigned long long)ss->id);
```
ggarber: isnt' "/" a valid character in the realm and username?  Shouldn't you escape it?

misi: True.
It is a general issue with the db backend.

bradleythughes: I hadn't considered that, actually. I just followed the pattern for the other Redis keys.

misi: I think I will merge this and open a new issue about the string escaping.

misi: @ggarber Many thanks again for pointing out the issue..

bradleythughes:
Thanks :)

Issue #211, opened 2018-02-09
Switch from Autotools to Cmake


RanjKuru:
It would be nice if somebody would help to switch to CMake.

firegurafiku:
@RanjKuru 
Are there official plans to switch to CMake or that's just your personal wishlist item?

misi:
Please send Pull Requests.. :-)

LorbusChris:
I'm interested in this as well in order to simplify my RPM spec for Fedora builds.
@RanjKuru have you done some work on this already? I'd happily review/finish a PR :)

LorbusChris:
This should include/fix:
- setting capabilities on dirs and files (and adding a --no-setcap build flag to disable this behaviour)
- fixing duplicate build IDs on turnserver and turnadmin

firegurafiku:
Okay, I'm ready to tackle this issue if nobody has already done significant progress. I also want to make use of [`FindPkgConfig.cmake`][1] for dependency resolution (see related issue #152). Any objections?

[1]: https://cmake.org/cmake/help/v3.0/module/FindPkgConfig.html

firegurafiku:
@LorbusChris 

Can you please explain what are the issues with capabilities and build IDs? How should CMake behave to deal with them properly?

LorbusChris:
@firegurafiku 
Here's the build log from my latest rpm build:
https://copr-be.cloud.fedoraproject.org/results/lorbus/coturn/fedora-27-x86_64/00722257-coturn/builder-live.log

I'm not sure what needs to change in order to deal with the duplicate build ids warning.

There are also multiple build warnings: 
[-Wimplicit-fallthrough=]
[-Wformat-truncation=]
[-Wstrict-prototypes]

In RPM lint there is an error which probably needs to be fixed/added in the code in this repo:
```
coturn.x86_64: E: missing-call-to-setgroups-before-setuid /usr/bin/turnadmin
coturn.x86_64: E: missing-call-to-setgroups-before-setuid /usr/bin/turnserver
```

Regarding the permission attributes on the files, I currently have to do that in the spec:
https://github.com/LorbusChris/coturn-rpm/blob/master/coturn.spec#L173

It would be nice to have a flag that optionally adds these during build, like e.g. here:
https://github.com/LibreOffice/online/blob/master/Makefile.am#L316 (also take a look at the configure file). I'd say this is not a requirement and could rather easily be added later.

see also: https://github.com/LorbusChris/coturn-rpm/issues/3

[...]

PR #216, opened 2018-02-20
Fix permissions on manpages and libturnclient

LorbusChris:
Fixes `spurious-executable-perm` warnings on manpages files and libturnclient during RPM linting.

This came up here: https://bugzilla.redhat.com/show_bug.cgi?id=1491492#c5

### DIFF
--- a/configure
+++ b/configure
@@ -559,7 +559,7 @@ if [ ${ER} -eq 0 ] ; then
 	INSTALL_MAN="ginstall -m 0644"
 	INSTALL_SCRIPT="ginstall"
 	INSTALL_SHARED_LIB="ginstall"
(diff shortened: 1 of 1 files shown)


LorbusChris:
Maybe all Install cmds shoould be reviewed to set permissions. Or is this maybe not wanted at all?
What do you think?

LorbusChris:
The attributes set on dirs and files during installation need an overhaul. This PR is not solving this.
You can see what permissions are needed on what files/dirs here: https://github.com/LorbusChris/coturn-rpm/blob/master/coturn.spec

Maybe setting capabilities should be optional (i.e. with --no-setcap configure flag).
Maybe this issue can be adressed when moving to cmake #211

PR #262, opened 2018-08-12
Update the example config file: use-auth-secret overrides lt-cred-mech.

krithin:
The comment before the TURN REST API flag was confusing: it seemed to imply that `use-auth-secret` _required_ the use of `lt-cred-mech`, but I've since learnt from PR #160 that in fact you're not supposed to use both of them together.

(I suspect that this was why lots of sample coturn config files you can find on the internet have both of those flags set.)

### DIFF
--- a/examples/etc/turnserver.conf
+++ b/examples/etc/turnserver.conf
@@ -180,8 +180,8 @@
 
 # TURN REST API flag.
 # Flag that sets a special authorization option that is based upon authentication secret.
-# This feature can be used with the long-term authentication mechanism, only.
-# This feature purpose is to support "TURN Server REST API", see
+# This feature cannot be used with the long-term credential mechanism.
+# This feature's purpose is to support "TURN Server REST API", see
 # "TURN REST API" link in the project's page 
 # https://github.com/coturn/coturn/
 #


misi:
If you use use-auth-secret then internally lt-cred-mech is automatically turned on.
So original text in config file I still think it is valid.

The problem is that I found earlier very confusing how these two configs behaves,
because if someone expect to set both than it will not allow to use both authentication in the same time!!!

use-auth-secret depends on lt-cred-mech, but if set then it overrides and does not allow to use the lt-cred-mech credentials anymore.

Sorry if I caused confusion with my warning in PR #160.

I am now confused about your fix / PR . 
Do you think that still we need to apply your update? I am not sure...

krithin:
(thanks for the quick review!)

If I understand your comment here correctly you're saying that the implementation of use-auth-secret in fact depends on the lt-cred mechanism internally.

However I think that from a user's perspective that doesn't matter: according to PR #160, it is an error to specify both use-auth-secret and lt-cred-mech in a config file, because you can only have one of the two turned on in a usable way.

That's why i sent this PR, to clear up the confusion about what goes in a correct config file; once this is through I can send PRs to other sample coturn configs (e.g. https://github.com/matrix-org/synapse/blob/master/docs/turn-howto.rst on the matrix project, which is what led me to coturn) to turn off lt-cred-mech in their sample configs too.

PR #264, opened 2018-08-16
Do not write to log before logging configuration is read from a config file

eiver:
find_config_file() now uses the "pass" variable to avoid printing "log file opened" to the log file during first pass of read_config_file() before logging is actually configured.

Moved print_features() behind both read_config_file() calls, so that the output goes to the configured log file.


misi:
Thanks!

PR #276, opened 2018-09-14
Added systemd service file example.

Liberasys:
Hello, I added a simple systemd service file example.
How to use it : 
put it in /usr/local/etc/
```
ln -s /usr/local/etc/coturn.service /etc/systemd/system/
systemctl daemon-reload
systemctl enable coturn
systemctl start coturn
systemctl status coturn
```

### DIFF
--- a/examples/etc/coturn.service
+++ b/examples/etc/coturn.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=coturn TURN Server
+Before=multi-user.target
+Before=graphical.target
+After=network-online.target
+After=remote-fs.target
+Wants=network-online.target
+
+[Service]
+ExecStart=/usr/local/bin/turnserver
+
+[Install]
+WantedBy=multi-user.target
+Alias=turnserver.service
+


misi:
Many Thanks! We have some similiar file in CentOS
https://github.com/coturn/coturn/blob/master/rpm/turnserver.service.fc
I want to merge and add privilidge drop etc.

misi:
@Liberasys @bradleythughes 
Would you mind if we don't merge this into the main repository but leave up to the packager to make a specific systemd service file. I would rather add to README some info about the various locations of different distributions systemd configs.
What do you think about it?  Is it Ok with you?
Or you still think it would be better to have a systemd service example in coTURN repository too?

Liberasys:
Hello,

I think that letting this job to the packages maintainer is an option.
Anyway, somebody that installs the last version from the repository (like me after having problems with  the debian stable version of coturn) should have all the information to get the service up. Maybe you could put some systemd example files in "example" folder, and add a note about how to install/enable them in the documentation ?

misi:
@Liberasys Ok. Sure. Thanks for the Quick response!
Now I am also try to help maintaining Debian package, so join us and help to make it better.. :)
Any help would appreciated..

Liberasys:
@misi : sorry, I am already running out of time :-) Good luck.

PR #284, opened 2018-10-02
added bandwidth usage reporting for bandwidth used by peers

AnsonKindred:
I needed to track all outgoing bandwidth so I expanded the bandwidth reporting to also report bandwidth used by peers. The traffic is reported separately from the client traffic so no existing functionality is changed. 

The redis keys for the peer data are: 
/turn/realm/[realm]/user/[user]/allocation/[allocation]/traffic/peer
/turn/realm/[realm]/user/[user]/allocation/[allocation]/total_traffic/peer
/turn/user/[user]/allocation/[allocation]/traffic/peer
/turn/user/[user]/allocation/[allocation]/total_traffic/peer

I tested with the turnutils_uclient and turnutils_peer for both tcp and udp connections and everything seems correct. I was not able to test passive tcp because I don't really know what it is and when I try and start the turnutils_uclient with -P I get errors.


misi:
Thanks! 
Could you please  add some documentation?
After it I will try to add to the next release..

misi:
Thanks!!

PR #356, opened 2019-03-04
Set and expire rather than publishing statdb

TaeMinPark:
**Why**
Coturn's statdb publish to Redis when the packets are turned. However, we want to set and expire that information in order to check if the user is still connected or not.

**Changes**
This PR set and expire for 5 seconds when the packets are turned.


TaeMinPark:
Sorry. Made PR to the wrong repository...

PR #359, opened 2019-03-08
Do not overload standard C API

bradleythughes:
There are many #defines in `src/ns_turn_defs.h` that simply define a new name for an existing, standard C API or type, sometimes with extra arguments (turn_realloc, turn_free). This adds an extra burden on programmers to remember to use the overloaded version instead of the standard one. Remove some of these overloads and refactor the code base to the standard APIs and types.

This PR also removes the currently unused pointer debugging facility. The rationale is that there are existing memory debugging tools and facilities in malloc()+friends implementations that we do not need to duplicate.


misi:
The question is why Oleg @mom040267 added all of these overloads.?
I guess there was a reason for it. 
Some exotic OS support? Any guess?

bradleythughes:
Not sure why, tbh. If I had to guess, I would say legacy, for example adding types of a specific size may have predated the int{8,16,32,64}_t types being introduced into the C standard (or needing to support older systems that didn't support new standards). I'm only guessing, of course, I could be totally wrong, but that isn't really the point. The point is to reduce the cognitive load while working on the code, especially when the overloaded types and API do nothing more than wrap the standard API.

misi:
My guess is code portability. 
I am not sure if windows or any other exotic platform has every standard C type and implemented the same way.. It should but I guess this own type definitions are a helper overlay. 
Maybe we don't need it anymore, maybe someone still needs it.

I see your point about simplifying the code, but I am not sure if it will not break some existing usages.

Anyhow the PR LGTM. 
So from my side let's merge it, and if someone is screaming about it, then we could revert it..

bradleythughes:
Thanks, @misi :) I don't think this will cause any problems, since I am just removing some indirection. We were using standard types already, just with our own invented names.

PR #367, opened 2019-03-18
Disconnect database gracefully

### LINKED ISSUE #366: turnadmin does not disconnect from DB properly
I use MySQL for userdb. When I add or delete user using turnadmin, MySQL reported notice like "Aborted connection [nnnnnnnn] todb: ...........". This causes accumulation of sockets to MySQL with TIME_WAIT.
In documents, using turnadmin is recommended for userdb management, but its not suitable for other application due to above problem.
We need documentation for userdb management from other application, especially creation of hmackey, or proper disconnection from DB for turnadmin.
My turnserver is v4.5.0.8.
BTW, I uses turnadmin from my Java application.

shu-mutou:
Coturn does not disconnect database properly when shutdown.
This behavior causes accumulation of sockets for database with TIME_WAIT.
Especially, `turnadmin` is recommended to manage userdb,
but use of `turnadmin` from other application or scripts
can run out of socket resource due to above problem.

This patch adds database disconnection when shutdown
for both of `turnserver` and `turnadmin`.

Fixes #366

### DIFF
--- a/src/apps/relay/dbdrivers/dbd_mongo.c
+++ b/src/apps/relay/dbdrivers/dbd_mongo.c
@@ -1349,6 +1349,15 @@ static int mongo_list_admin_users(int no_print)
 	return ret;
 }
 
+static void mongo_disconnect(void) {
+	MONGO * mongoconnection = (MONGO *) pthread_getspecific(connection_key);
+	if (mongoconnection) {
+		MongoFree(mongoconnection);
+		mongoconnection = NULL;
+	}
+	TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "MongoDB connection was closed.\n");
+}
+
 //////////////////////////////////////////////////////////
 
 static const turn_dbdriver_t driver = {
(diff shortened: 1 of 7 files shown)


misi:
Thanks! Seems Ok..
Have you tested with all type of db backends?

shu-mutou:
Sorry, I could test only for mysql. I don't have other backends.

shu-mutou:
BTW, could you let me know backport policy? Would this PR be backported to stable versions?

misi:
Unfortunately due the limited resources I don't know yet about any backport policy. 
AFAIU we include to new releases (master branch) both the new features and the fixes..
If you need to backported fix then I propose to use your favorite distribution patch system and backport a patch actually. 
If you think this mechanism could not work for you, somehow it is impossible or not suitable, then let me know and we can discuss it further.

shu-mutou:
OK, I see. Thank you for your reply. 😄
I can setup coturn with my patch for our service from my forked repository, so there is no need to rush backport. However, I'm happy if important fixes would be backported into latest stable version. 😉

misi:
It would be great if you (and many other coturn devops) could join as developer and then we could create new branch per release from the master. 
 ;-)
I fully agree that it would the right way..

misi:
@shu-mutou  Many Thanks for your patience and I am so sorry for the long delay! Thanks again!

PR #392, opened 2019-05-14
Wrong fread usage

islamoglus:
Hi,

Return value of fread function is not zero on success therefore I modified these sections.

Check:
The paragraph shown below quoted from: http://man7.org/linux/man-pages/man3/fwrite.3.html
RETURN VALUE 
       On success, fread() and fwrite() return the number of items read or
       written.  This number equals the number of bytes transferred only
       when size is 1.  If an error occurs, or the end of the file is
       reached, the return value is a short item count (or zero).

Thanks for review
Selahaddin

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -1693,6 +1693,7 @@ static void read_config_file(int argc, char **argv, int pass)
 static int adminmain(int argc, char **argv)
 {
 	int c = 0;
+	int rc = 0;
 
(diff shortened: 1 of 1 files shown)


misi:
@islamoglus 
I think it is valid issue what you pointed out.  We need to add another check.
But there are two case
1.  if it is 0 
2.  if it is not as long as we expected

islamoglus:
Updated to check both states

misi:
Many Thanks! & I am so sorry for my late reply!!

PR #398, opened 2019-05-24
Allow MD5 in FIPS mode

byronclark:
This is one of those special cases where a non approved cryptographic
algorithm is allowed when operating in FIPS mode. Inform OpenSSL that
this is the case.

In the STUN RFC the long-term credential mechanism requires that the key
used in the HMAC-SHA1 generation be the MD5 of specific values:
https://tools.ietf.org/html/rfc5389#section-15.4

Since this is obfuscating parameters to be used in an approved
cryptographic algorithm, this is allowed usage per the [FIPS 140-2 Implementation Guidance](https://csrc.nist.gov/csrc/media/projects/cryptographic-module-validation-program/documents/fips140-2/fips1402ig.pdf).
See page 81.

Without this change, coturn crashes when trying to set up any long-term
credential mechanism.


misi:
Thanks! I will review it..

PR #417, opened 2019-07-26
Append only to log files rather to override them

robert-scheck:


misi:
Thanks! I will review it..

PR #442, opened 2019-09-27
Updated incorrect string length check for 'ssh'

chanduthedev:
ssh length should be 3, but checking for 4. So updated memcmp method to 3 instead of 4.


misi:
Thanks! I will merge it.

chanduthedev:
Thank you for the response.

Issue #426, opened 2019-08-08
Error while docker build coturn


thachnv92:
Hello,
I got error when build coturn in docker/coturn/
In Step 18/26:
```
Step 18/26 : RUN export DEBIAN_FRONTEND=noninteractive &&       apt-get update &&       apt-get install -y libc6>=2.15 libevent-core-2.0-5>=2.0.10-stable libevent-extra-2.0-5>=2.0.10-stable libevent-openssl-2.0-5>=2.0.10-stable libevent-pthreads-2.0-5>=2.0.10-stable libhiredis0.13>=0.13.1 libmariadbclient18>=5.5.36 libpq5>=8.4~ libsqlite3-0>=3.6.0 libssl1.1>=1.1.0 libmongoc-1.0 libbson-1.0
 ---> Running in de215bb3e855
Get:1 http://security-cdn.debian.org/debian-security stable/updates InRelease [39.1 kB]
Get:2 http://cdn-fastly.deb.debian.org/debian stable InRelease [118 kB]
Get:3 http://security-cdn.debian.org/debian-security stable/updates/main amd64 Packages [53.4 kB]
Get:4 http://cdn-fastly.deb.debian.org/debian stable-updates InRelease [46.8 kB]
Get:5 http://cdn-fastly.deb.debian.org/debian stable/main amd64 Packages [7897 kB]
Fetched 8154 kB in 2s (4773 kB/s)
Reading package lists...
E: Unable to locate package libevent-core-2.0-5
E: Couldn't find any package by glob 'libevent-core-2.0-5'
E: Couldn't find any package by regex 'libevent-core-2.0-5'
E: Unable to locate package libevent-extra-2.0-5
E: Couldn't find any package by glob 'libevent-extra-2.0-5'
E: Couldn't find any package by regex 'libevent-extra-2.0-5'
E: Unable to locate package libevent-openssl-2.0-5
E: Couldn't find any package by glob 'libevent-openssl-2.0-5'
E: Couldn't find any package by regex 'libevent-openssl-2.0-5'
E: Unable to locate package libevent-pthreads-2.0-5
E: Couldn't find any package by glob 'libevent-pthreads-2.0-5'
E: Couldn't find any package by regex 'libevent-pthreads-2.0-5'
E: Unable to locate package libhiredis0.13
E: Couldn't find any package by glob 'libhiredis0.13'
E: Couldn't find any package by regex 'libhiredis0.13'
E: Package 'libmariadbclient18' has no installation candidate
[...]

PR #449, opened 2019-10-07
Fix Dockerfile for latest Debian

rao-donut:
Hi,

An error occurred in the docker build for coturn.
So I changed the Dockerfile to build succeeds in the latest Debian.

Fix for #426 .


misi:
Many  Thanks!
I will check it.

PR #470, opened 2019-11-20
Compiler warning fixes

bobsayshilol:
These fix a few issues when compiling coturn with high level warnings enabled as well as allowing users to pass constant data as parameters to (some) functions that do not mutate them.


bobsayshilol:
Thanks!

PR #471, opened 2019-11-20
Fix a memory leak when an SHATYPE isn't supported

bobsayshilol:
Deallocating `str` happens at the end of the function, so don't skip it when encountering an unsupported `SHATYPE`.


bobsayshilol:
Thanks!

PR #472, opened 2019-11-20
Add additional input validation for STUN messages

bobsayshilol:
The STUN attribute API wasn't checking that the size the buffer reported itself to be was not greater than the size of the buffer available, and `stun_attr_ref`s were not size-checked before being returned to the caller.

### DIFF
--- a/src/client/ns_turn_msg.c
+++ b/src/client/ns_turn_msg.c
@@ -360,7 +360,14 @@ int stun_get_command_message_len_str(const uint8_t* buf, size_t len)
 {
 	if (len < STUN_HEADER_LENGTH)
 		return -1;
-	return (int) (nswap16(((const uint16_t*)(buf))[1]) + STUN_HEADER_LENGTH);
+
+	/* Validate the size the buffer claims to be */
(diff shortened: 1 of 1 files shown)


misi:
If I am correct Memory in different architectures is no always 4 byte aligned. 
You use 4 byte alignment as constant. Are you sure it will work on all kind of architecture your change?

bobsayshilol:
Which line are you referring to? I don't see any changes from aligned to unaligned loads/stores but I could be missing something. There's also a few `((const uint32_t*)(buf))[1]`'s in that file as well as other similar casts for `uint16_t`'s so I assumed alignment wasn't a concern.

misi:
@bobsayshilol 
Sorry I misunderstood your comment in the code.  `/* Round to boundary */`
If I understand correctly you mean the 32bit stun boundary alignment.
Sorry..

bobsayshilol:
Ah I see. Yeah that's correct, it's for rounding up the size of a STUN packet using the same approach as `stun_attr_get_next_str()`.

And thanks for merging!

PR #478, opened 2019-12-19
turnserver.conf: add --prod section to enhance security

alepolidori:
As reported by [RFC:](https://tools.ietf.org/html/rfc5389#section-16.1.2)
```
Revealing the specific software version of the agent through the
SOFTWARE attribute might allow them to become more vulnerable to
attacks against software that is known to contain security holes.
Implementers SHOULD make usage of the SOFTWARE attribute a
configurable option.
```
Just add `prod` to the configuration file to hide software version in Binding Success Response

image

### DIFF
--- a/docker/coturn/turnserver.conf
+++ b/docker/coturn/turnserver.conf
@@ -575,6 +575,15 @@ syslog
 #
 #stun-only
 
+# Option to hide software version. Enhance security when used in production.
+# Revealing the specific software version of the agent through the
+# SOFTWARE attribute might allow them to become more vulnerable to
+# attacks against software that is known to contain security holes.
+# Implementers SHOULD make usage of the SOFTWARE attribute a
+# configurable option (https://tools.ietf.org/html/rfc5389#section-16.1.2)
+#
+#prod
+
 # Option to suppress STUN functionality, only TURN requests will be processed.
 # Run as TURN server only, all STUN requests will be ignored.
 # By default, this option is NOT set.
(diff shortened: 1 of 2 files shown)


dbrgn:
Is the name "prod" a good choice? Will there be other behavior controlled by this option in the future?

If yes, this should be documented as "Enable production mode". If it's just about hiding the software version, then maybe it should be renamed to "no-software-version" or "no-software-attribute" before the next release?

(Especially people coming from tools like NodeJS might expect "prod mode" to do more than just hiding the version.)

I'm happy to provide a PR if you agree.

misi:
I agree. Send a PR..

dbrgn:
Ah, I didn't realize that the "--prod" command line option was already released a few years ago. In that case it cannot be renamed without breaking anything, so I deprecated it instead (it still works though). See #506.

matiaslopezd:
So is possible to hide the software attribute with `no-software-attribute` in config?

dbrgn:
You're on a version from 2019 that does not yet support the renamed `no-software-attribute`. Adding `prod` to the config or specifying `--prod` in the launch arguments should work though, I think.

dbrgn:
Ah, never mind, as you can see this PR was merged on Feb 12, 2020, so your version is too old. You could ask the Ubuntu maintainers to backport the `no-software-attribute` fix, but I doubt they'll do that.

matiaslopezd:
Mmmh so, since which version is possible to use `no-software-attribute`? I check [here](https://launchpad.net/ubuntu/+source/coturn) and I can update to the latest version 4.5.1.3-1.

dbrgn:
I think you can use `no-software-attribute` starting with 4.5.1.2 (according to the changelog).

PR #481, opened 2019-12-27
Dbd rest and remote_auth_api usage

ovoshlook:
added dbd_rest interface with remote_auth_api  usage option possibility
allows to connect to 3-d party server via RESTfull API of that server and fetch secret based on username

-- not sure if needed for now. but better remove after review
oauth functionality via RESTfull API of 3-d party server but not supported officially.
docker file updated with debian-stretch as well

I'm not sure about usage of 'W' and 'w' keys and about couple of of other things. So suggestions are required.


misi:
Many Thanks for the PR. 
The name is confusing little bit, because for oauth already used the RESTAPI name. 
Maybe we should refer this as http backend or curl backend? What do you think?
OAuth has one big benefits over the apporoach that the client and server operated by the same entity. The operator of the turn service and the ICE /Turn client application could be separated nicely. So I don't think oauth should be replaced. 

Some more documentation needed in README, and updated man page (run make-man.sh to create from the README  man page..)

ovoshlook:
Sure! I Will update it during the weekend.
Http_backend sounds nice. I Will try do add more clear descriprion.

ovoshlook:
hi sorry for the loooong reply... Had no time to do it earlier.
We made couple of changes -plz check it when you will have time  while I will resolve conflicts

ovoshlook:
I will close this one as I am accidentally added new[ PR](https://github.com/coturn/coturn/pull/524) with the same code that also updated during our discussions here and added some new configuration control mechanisms. That one is more fresh.

PR #506, opened 2020-03-13
Rename "prod" config option to "no-software-attribute"

dbrgn:
As discussed in #478, if the parameter only controls whether or not to send the software attribute and not other production-relevant configurations, it should be named accordingly.

The old --prod configuration option still works, but is now deprecated and undocumented.

### DIFF
--- a/README.turnserver
+++ b/README.turnserver
@@ -158,7 +158,7 @@ Flags:
 
 -o, --daemon		Run server as daemon.
 
---prod       	 	Production mode: hide the software version.
+--no-software-attribute	Production mode: hide the software version.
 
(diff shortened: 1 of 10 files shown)


Werkov:
Hi, I noticed the version string in network dump and I was surprised that it's punched inside every UDP datagram (this was IP telephony application) whose size ranged from 138B to 638B and the version string itself was 26B. I'd say that's non-negligible overhead.

I suggest making the logic inverted, i.e. don't punch version by default and make it visible only with an option (that should still be RFC-compatible behavior).

(Also, more or less related, what about dropping the attribute completely? That's at least another 4B/datagram less :-))

PR #514, opened 2020-03-30
Unify spelling of Coturn

paulmenzel:


misi:
Thanks!

misi:
I like more to use coTURN, but Oleg the origin author, and 99% code owner used Coturn. 
I agree that it is not nice to have different names if we speak about the same thing.

Thanks again..

PR #516, opened 2020-04-02
Update docker-compose-all.yml

raghumuppa:


raghumuppa:
The volume path of mysql and porstgres has to be change.

raghumuppa:
updated mysql, postgres volume paths in docker-compose-all.yaml file.

misi:
@raghumuppa Many Thanks for your patience!

PR #522, opened 2020-04-13
Drop of supplementary group IDs

weberhofer:
Fix related to POS36-C and rpmlint error "missing-call-to-setgroups-before-setuid".


misi:
Thanks!

PR #524, opened 2020-04-15
backend-http-client and dbd_rest

ovoshlook:
added dbd_rest interface with remote_auth_api usage option possibility
allows to connect to 3-d party server via REST of that server and fetch secret based on username

-- not sure if needed for now. but better remove after review
oauth functionality via RESTfull API of 3-d party server but not supported officially.
docker file updated with debian-stretch as well

I'm not sure about usage of 'W' and 'w' keys and about couple of of other things. So suggestions are required.
automatic switching off REST API built it in coturn if remote auth api switched on.


misi:
Many Thanks!
I will merge it in the next version 4.5.2.x

ovoshlook:
@misi Hi
Is there is a way it will be merged one day?

Neustradamus:
@misi: Have you seen?

misi:
@ovoshlook  Yes sorry for the delay I will try to integrate it ASAP.

misi:
@ovoshlook I started to check your PR, rebased on master. Fixed small issues. 
I don't have enough info in the docs to understand what is the primary use case. Maybe you should add more example..
I see some trace that it should be use with kamailio db?

I am not quite happy how your backend works to get credential one by one from the rest api.
It is very different how the other database backends works. They load the credentials tables in one step and async io make the server fast and non blocking.
AFAIU your solution do it very different way. It could be very dangerous, that it may blocking all the things. Have you tested it on high load?

misi:
@ovoshlook  Other issue I have found that get_auth_secret don't have initializer.

misi:
@ovoshlook To use coturn with kamailio I am prerrty sure you can use one mysql server where you can create some table views or replication table with triggers, or sync the two with event scheduler.

misi:
@ovoshlook All in all I feel  your PR is not yet ready to merge in the next release. We need further investigate if it is blocking or not the async io threads.

misi:
@ovoshlook  see the branch PR524 on top of actual master.

ovoshlook:
Hello
sorry for the late reply


> AFAIU your solution do it very different way. It could be very dangerous, that it may blocking all the things. Have you tested it on high load?
Yes It runs on a high load. 
No problems so far.

The main reason for this solution is to create yet another interface unrelated to databases, which allows services to talk directly between each other. So it is not specially Kamailio based solution. Kamailio in that case just an example.

> @ovoshlook To use coturn with kamailio I am prerrty sure you can use one mysql server where you can create some table views or replication table with triggers, or sync the two with event scheduler.
> It is very different how the other database backends works. They load the credentials tables in one step and async io make the server fast and non blocking.

It is not about a possibility of the Kamailio to use databases but it is about having a possibility 3d service decide what to use and abstract in that way coturn itself from requesting data from a particular database by using a more generic REST interface. Also here is another possibility opened to not overload the database if one-time passwords generated by 3d party service. With this solution in my case, I decreased the number of requests to the database proportionally to the number of instances of 3d service because 3d service decides for now does it needed to request DB or not. 
For example:
- 3d service generates username/pass pair
- Sends it to the client
- Client requests connection to Coturn with those credentials
- Coturn requests service if credential exists
- 3d party service responses based on the existence
If I would use just a shared DB between Coturn and 3d service I would need to go to DB every time because the cache of Coturn would never have requested username:password pair as it is onetime.

That is why it works different ( not uses domain for the fetching username-pass pairs but only one per request username/pass)

I would check one more time about blocking operations because I implemented in the same way how db_redis and db_mysql works. It uses **dbdriver** abstraction as you can see so it calls **db_rest** connector in the same way. As I see here is nonblocking implemented by using pthreads (DB connections work in a different thread than relay connections), so should work in the same way then.


But, I agree that **db_rest** driver has to be implemented closer to the way with other connections for not to give some specific deviation for the whole system itself, as my implemented case is just a specific case that can be implemented in the same logical way other connectors implemented:
- fetch all and save to the cache, 
- check by request from the client if no username in the cache
- go to 3d service to request if this username exists


> @ovoshlook Other issue I have found that get_auth_secret don't have initializer.

I will fix this one and will add the logic to make this interface work closer to other db_connectors and then update the rest of the code.

Thank you for your comments and questions. It shows how my explanation and docs were messed about understanding what is it for. If something unclear in my explanation in this comment - plz let me know, I will answer it as fast as possible.

misi:
I need more time to test and  think about it. Sorry for the delay. I hope you understand that I can't merge it until I double-check if it will not break anything.

ovoshlook:
I think your point to make this interface work closer ( to not fetch particularly one by one by username but by domain, because it still available to fetch it one by one then but not change the template ) to the rest of the interfaces is pretty clear and correct.
So I believe you won't merge it into master before I'll make this driver behaviour closer to the rest of the interfaces.

I just explained above the reason to make this interface.
Let me make it more generic then and after you will be able to recheck and test.

misi:
I have limited time that I split between eduMeet and coturn project so if I  answer only slowly it is beacuse I am fragmented. 
I need bigger time window to test and check your code, but I definitely want to merge if it does not make any trouble. It behaves very differently and it couldn't be as effective as reading credntial data to memory and search in the memory, and verify data based on that. To send out after every TURN auth a  new backend REST request and (may block execution further until rest comes back), is somehow unusual, and we need to be very careful and warn users about penalty even if it works fine.
So I will try to do my best to find time for it.

Neustradamus:
@misi: Happy New Year 2022!

Any news on this PR?

Thanks in advance.

ggarber:
@ovoshlook Are you still interested on the work in this branch?

ovoshlook:
@ggarber 
Seems not ( in nearest future ).

ggarber:
> @ggarber Seems not ( in nearest future ).
Ok, then i'll close it for now and you or anybody interested on this work can reopen it later on.

Thank you very much for the work here, I think it is an interesting feature to be considered in the future for coturn 🙏

PR #525, opened 2020-04-18
Add flags to disable periodic use of dynamic tables

gfodor:
We are using coturn in an environment which supports database "pausing", whereby the database is taken offline to minimize costs when there is no underlying SQL activity. (In our case, this is with AWS Aurora Serverless.)

In coturn, there are three periodic requests being made to the database, which we would like to disable to ensure the database is not forcibly kept online when using coturn:

- Periodic health check pings to the `turn_secret` table (this seems to be exclusively informational to let site operator know the database is inaccessible)
- Periodic syncs of allow/denied peer ip lists
- Periodic syncs of dynamic realm origins + options

This PR adds 3 options which can be used to selectively disable these:

- `--no-auth-pings`
- `--no-dynamic-ip-list`
- `--no-dynamic-realms`

These options are also useful because by disabling them if not needed, you can avoid creating the relevant tables in your database.


misi:
Thanks for the PR.
I will merge it, just add new option to README.turnserver
 and run make-man.sh to update man pages too

misi:
@gfodor I want to release 5.1.2 today..

gfodor:
OK! Done, please verify it looks right, I haven't edited man pages before so am not totally sure if things are being reflected properly.

misi:
Thanks!

Issue #220, opened 2018-03-06
Add option to disable SSLv3


donlorenzo:
By default coturn seems to support SSLv3, TLSv1, TLSv1.1, TLSv1.2.
There are options to disable the newer protocol versions ("--no-tlsv1", "--no-tlsv1_1", and "--no-tlsv1_2") but there does not seem to be an option to disable to old SSLv3 ("--no-ssl23" ?).

It would be nice to have than option to selectively disable support for SSLv3.

bradleythughes:
SSLv3 (and SSLv2) are unconditionally disabled by default. You can see that in the code here:
https://github.com/coturn/coturn/blob/master/src/apps/relay/mainrelay.c#L2642-L2650

According to the changelog, SSLv2 was removed in 4.4.1.1, and SSLv3 in 4.5.0.3:
https://github.com/coturn/coturn/blob/master/ChangeLog#L133
https://github.com/coturn/coturn/blob/master/ChangeLog#L44-L48

donlorenzo:
Thanks for the reply.
I guess what surprised me (and still does) is that when I start coturn it reports loading certificate and private key for SSL23:

> 0: SSL23: Certificate file found: *snip*
> 0: SSL23: Private key file found: *snip*
> 0: TLS1.2: Certificate file found: *snip*
> ...

Before I disabled TLSv1 and v1.1 coturn also reported loading of keys for those protocols. Therefore I assumed that it still supports SSL23 since it is reporting the keys being loaded.
Also [these lines](https://github.com/coturn/coturn/blob/master/src/apps/relay/mainrelay.c#L2721-L2722) made me think that SSL was supported for "compatibility".

bradleythughes:
Indeed, that does seem confusing. I do not quite understand why it creates that context, though...

MichaIng:
Vote to get this confusion fixed: https://help.nextcloud.com/t/howto-setup-nextcloud-talk-with-turn-server/30794/41?u=michaing
🙂

hucste:
i vote too. 
This confusion should not exists! ;)
It's not normal to display a text about all versions SSL if are disabled.

eakraly:
As of today, openssl-1.1.1 and above have SSLv3 explicitly disabled (need to recompile openssl to enable it)

PR #566, opened 2020-05-29
Remove reference to SSLv3

bpcurse:
Judging from the information provided here https://github.com/coturn/coturn/issues/220#issuecomment-371916988,
SSL isn't supported anymore so the reference should be removed from the config file.


misi:
Thanks!

PR #574, opened 2020-06-09
Change Docker turnserver.conf to latest turnserver.conf

DevRockstarZ:
Sorry for wrong PRs.


misi:
Many Thanks and sorry for the late reply!

benbro:
Why not call it coturn.conf?

Issue #556, opened 2020-05-22
Centos 7 RPM build


omaximenco:
./CentOS7.pre.build.sh
./build.sh

Getting this error:
RPM build errors:
    Installed (but unpackaged) file(s) found:
   /usr/share/turnserver/etc/cacert.pem
   /usr/share/turnserver/etc/coturn.service

4.5.1.1 works fine.

misi:
It is fixed so I close it! Thanks for your contribution!

PR #575, opened 2020-06-09
fix for #556

osterik:


misi:
Many Thanks! Sorry for the late reply

PR #577, opened 2020-06-09
build RPM from local folder instead of git repo

osterik:
Hello! 

RPM package build script always uses the master git repo, so it makes it hard to add local patches.
I had faced up with issue #556 and was surprised that my attempts to fix it ignored. So, I had to make this patch to have an ability to resolve the RPM-issue.

### DIFF
--- a/rpm/build.settings.sh
+++ b/rpm/build.settings.sh
@@ -5,7 +5,6 @@
 TURNVERSION=4.5.1.2
 BUILDDIR=~/rpmbuild
 ARCH=`uname -p`
-TURNSERVER_GIT_URL=https://github.com/coturn/coturn.git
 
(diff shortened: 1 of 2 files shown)


misi:
@osterik 
What is the use case to package local version? I am not sure that we should make this change.
Why it is not good to pull a release from git? I feel safe to pull it from git.
Thanks!

ilya-kisleyko-oxagile:
The main goal of this patch - have the ability to build your own package with local changes, instead of using upstream.

osterik:
The main goal of this patch - have the ability to build your own package with local changes, instead of using upstream.

misi:
Normally we package upstream version, but I do not have too much experience in rpm packaging, so I don't have strong opinion on it.

PR #579, opened 2020-06-11
Ignore MD5 for BoringSSL

islamoglus:
Hi,

These lines added to allow MD5 for OpenSSL in FIPS mode previously but BoringSSL does not allow MD5 in FIPS mode. So, I added definition check to chek MD5 support of SSL library for BoringSSL compatibility.

Thanks for reviewing.
Selahaddin.


misi:
Thanks! And sorry for the late reply!

PR #599, opened 2020-07-19
Do not use FIPS and remove hardcode OPENSSL_VERSION_NUMBER with LibreSSL

### LINKED ISSUE #552: LibreSSL compilation failures
Compilation with a current version of LibreSSL produces an error and a warning:

- [LibreSSL has no FIPS mode](https://marc.info/?l=openbsd-misc&m=139819485423701) so `FIPS_mode()` is not defined:
https://github.com/coturn/coturn/blob/113f13855a53aeb41ca5b71aa36b3f945cee2052/src/client/ns_turn_msg.c#L247-L249
https://github.com/coturn/coturn/blob/113f13855a53aeb41ca5b71aa36b3f945cee2052/src/client/ns_turn_msg.c#L257-L259

  A potential fix is to enclose both calls with `#ifdef OPENSSL_FIPS ... #endif` which was removed in https://github.com/coturn/coturn/commit/0e03fa86df3c2ba3afb3793c724d3579afd1ecf3 .

- Hardcoding an old OpenSSL version number if LibreSSL is detected
https://github.com/coturn/coturn/blob/113f13855a53aeb41ca5b71aa36b3f945cee2052/src/apps/common/ns_turn_openssl.h#L45-L48
causes a problem when compiling
https://github.com/coturn/coturn/blob/113f13855a53aeb41ca5b71aa36b3f945cee2052/src/apps/relay/dtls_listener.c#L209-L211 as newer LibreSSL versions need the `const` modifier.

  To address this either revisit the mapping of LibreSSL to an old OpenSSL version number or use an explicit `| defined(LIBRESSL_VERSION_NUMBER)` here.

cedk:
Fix #552

### DIFF
--- a/src/apps/common/ns_turn_openssl.h
+++ b/src/apps/common/ns_turn_openssl.h
@@ -42,9 +42,4 @@
 #include <openssl/dh.h>
 #include <openssl/bn.h>
 
-#if (defined LIBRESSL_VERSION_NUMBER && OPENSSL_VERSION_NUMBER == 0x20000000L)
-#undef OPENSSL_VERSION_NUMBER
-#define OPENSSL_VERSION_NUMBER 0x1000107FL
-#endif
-
 #endif //__NST_OPENSSL_LIB__
--- a/src/apps/relay/ns_ioalib_engine_impl.c
+++ b/src/apps/relay/ns_ioalib_engine_impl.c
@@ -1828,7 +1828,7 @@ int ssl_read(evutil_socket_t fd, SSL* ssl, ioa_network_buffer_handle nbh, int ve
 	BIO* rbio = BIO_new_mem_buf(buffer, old_buffer_len);
 	BIO_set_mem_eof_return(rbio, -1);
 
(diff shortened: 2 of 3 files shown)


Neustradamus:
@misi: Have you seen?

misi:
@cedk  Do you think removing this default ssl version value will not cause any issue? I guess the reason it was there somehow to fallback if it is not configured. What do you think?

cedk:
For me it prevents to use other feature of libressl to hard code its openssl version.
It is better to not use a feature based on the version of the library than pretending that a library has a different version.
As far as I have seen other codes use similar test when dealing with openssl and libressl compatibility.

misi:
I am not expert on this field, so I hope it will not make any trouble.. :)
Thanks for your contribution!

Neustradamus:
@misi: Thanks for merging of several PRs.

Please look other current PRs, last comments/changes and please do a new comment if it is not good...

misi:
@Neustradamus I do my best as always. :-)

PR #618, opened 2020-08-25
Print full date and time in logs

PaulWay:
The 'number of seconds since the daemon was started' is amusing and compact, but it hinders debugging and correlation with other events.  This implements logs that use a complete `YYYY-MM-DDTHH:MM:SS` output which is easy to read and correlate.

Signed-off-by: Paul Wayper <paulway@mabula.net>

### DIFF
--- a/src/apps/common/ns_turn_utils.c
+++ b/src/apps/common/ns_turn_utils.c
@@ -159,7 +159,7 @@ void set_no_stdout_log(int val)
 }
 
 #define MAX_LOG_TIMESTAMP_FORMAT_LEN 48
-static char turn_log_timestamp_format[MAX_LOG_TIMESTAMP_FORMAT_LEN] = "%Y-%m-%dT%H:%M:%S";
+static char turn_log_timestamp_format[MAX_LOG_TIMESTAMP_FORMAT_LEN] = "%FT%T%z";
 
 void set_turn_log_timestamp_format(char* new_format)
 {
--- a/src/apps/common/ns_turn_utils.c
+++ b/src/apps/common/ns_turn_utils.c
@@ -178,8 +178,8 @@ void turn_log_func_default(TURN_LOG_LEVEL level, const char* format, ...)
 #define MAX_RTPPRINTF_BUFFER_SIZE (1024)
 		char s[MAX_RTPPRINTF_BUFFER_SIZE+1];
 #undef MAX_RTPPRINTF_BUFFER_SIZE
-		struct tm local_now = localtime(time(NULL));
-		strptime(s, sizeof(s)-100, "%Y-%m-%dT%H:%M:%S", &local_now);
+		time_t now = time(NULL);
+		strftime(s, sizeof(s), "%Y-%m-%dT%H:%M:%S", localtime(&now));
 		snprintf(s + 19,sizeof(s)-100,(level == TURN_LOG_LEVEL_ERROR) ? ": ERROR: " : ": ");
 		size_t slen = strlen(s);
 		vsnprintf(s+slen,sizeof(s)-slen-1,format, args);
--- a/src/apps/common/ns_turn_utils.c
+++ b/src/apps/common/ns_turn_utils.c
@@ -158,6 +158,8 @@ void set_no_stdout_log(int val)
 	no_stdout_log = val;
 }
 
+int use_new_log_timestamp_format = 0;
+
 void turn_log_func_default(TURN_LOG_LEVEL level, const char* format, ...)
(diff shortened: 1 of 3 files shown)


PaulWay:
Haven't checked if this compiles or not yet, and haven't removed the old `log_start_time` and `log_time()` code.

One possibility I though of was implementing this as an option that could be turned on or off, say by a command-line or configuration option.  Haven't investigated how to do that yet.

misi:
Please add flag / config option for that to avoid breaking changes

PaulWay:
OK - do you want a flag to turn on the new log format, or a flag to turn on the old log format?  I.e. which log format should be the preferred?  I'm curious to know what value you see in the old format?

I'm happy to avoid a 'breaking' change in a significant feature or piece of functionality, but I'm not sure I see the utility of the old functionality.  I suppose I see it as a bug to be fixed rather than a feature that needs to be preserved.  Help me understand the significance of the old relative timestamp.

PaulWay:
And would you prefer that flag, e.g. `int use_new_log_timestamp format`, to be in `apps/common/ns_turn_utils.h` or somewhere else?

PaulWay:
This whole log printing is really puzzling me.

There's two separate places which format log lines.  One is in `turn_log_func_default`, which then calls the other in `vrtpprintf`.  This seems to do the whole process of reading the varargs and formatting the log line twice, before printing it to different locations (stdout in `turn_log_func_default`, syslog or the log file in `vrtpprintf`).  This seems to waste time doing the same job twice.

Is there a particular reason it's done this way?

What I'd propose to do here is to format the log line once and then print that to stdout, the log file and syslog as required by the relevant flags.  This would all be handled in the one function `turn_log_func_default`, with the `vrtpprintf` function being effectively folded into it.  One buffer would be allocated to format the log line, and this would reduce the code size and give more consistency.

But I won't go ahead with that if there's strong opposition to it, or if there's a particular reason for doing it the current way which I don't understand.  So please let me know what you'd prefer me to do here.

PaulWay:
So I've merged the log printing into one function, and allowed users to set the timestamp as well.  Still trying to test the code to see if it actually works.

PaulWay:
@misi Any thoughts?

misi:
@PaulWay  Sorry for my late reply!
> OK - do you want a flag to turn on the new log format, or a flag to turn on the old log format? I.e. which log format should be the preferred? I'm curious to know what value you see in the old format?

I want a flag for the new to be backward compatible.
> 
> I'm happy to avoid a 'breaking' change in a significant feature or piece of functionality, but I'm not sure I see the utility of the old functionality. I suppose I see it as a bug to be fixed rather than a feature that needs to be preserved. Help me understand the significance of the old relative timestamp.

I believe we need smooth transition. It could be someone built systems on it, so I propose to have new flag and indicate in the docs that it will flipped in the next version and your will be the default, and after one version iteration the default behavior will change to the new, to give time to prepare for the change.

misi:
> This whole log printing is really puzzling me.
> 
> There's two separate places which format log lines. One is in `turn_log_func_default`, which then calls the other in `vrtpprintf`. This seems to do the whole process of reading the varargs and formatting the log line twice, before printing it to different locations (stdout in `turn_log_func_default`, syslog or the log file in `vrtpprintf`). This seems to waste time doing the same job twice.
> 
> Is there a particular reason it's done this way?
> 
> What I'd propose to do here is to format the log line once and then print that to stdout, the log file and syslog as required by the relevant flags. This would all be handled in the one function `turn_log_func_default`, with the `vrtpprintf` function being effectively folded into it. One buffer would be allocated to format the log line, and this would reduce the code size and give more consistency.
> 
> But I won't go ahead with that if there's strong opposition to it, or if there's a particular reason for doing it the current way which I don't understand. So please let me know what you'd prefer me to do here.

Unfortunately I am not the original author of the code, so I can't help in it why it was designed that way. We can simplify, but need to be backward compatible and not break any usage. So let's make it in a different PR..

misi:
> @misi Any thoughts?
@PaulWay 
Sorry I have limited time.. :-(
Accept my apologies that I can only reach here now..

misi:
We need to also think that wide OS supported so it should work and compatible with more exotic OSs like mac, win, etc.

PaulWay:
Hi @misi - thanks for that, sorry for not getting back to you earlier.  Hope this all works well!

misi:
@PaulWay I adjusted little bit (add doc, fix cli etc.) but it seems working fine now. Thanks for your help and contribution!

PR #637, opened 2020-10-17
Delete trailing whitespace in example configuration files

edwargix:
FYI, this patch was created by doing `M-x delete-trailing-whitespace` in GNU Emacs.


misi:
Thanks!

PR #643, opened 2020-10-30
Allow RFC6062 TCP relay data to look like TLS; fixes #605

### LINKED ISSUE #605: tcp_peer_accept_connection: cannot set TCP peer data input callback
With some frequency when a peer tries to connect to an allocated port the peer gets a connection reset and I see this in the coturn log

3118: IPv4. tcp accepted from: 64.99.230.16:50771
3118: tcp_peer_accept_connection: cannot set TCP peer data input callback

Typically the peer might have to try connecting a couple more times before getting a connection. 

Can anybody shed some light on the meaning of this error message and how I might prevent it from happening.

tupelo-schneck:
This point in the code was making a best-effort attempt to reject TLS connections on non-TLS listeners.

After this is merged, it will skip this check when the socket is being used for RFC6062 TCP relay data.  That data should be opaque for the TURN server and it is fine if it happens to be (or look like) TLS.

Fixes #605 .

### DIFF
--- a/src/apps/relay/ns_ioalib_engine_impl.c
+++ b/src/apps/relay/ns_ioalib_engine_impl.c
@@ -3374,7 +3374,7 @@ int register_callback_on_ioa_socket(ioa_engine_handle e, ioa_socket_handle s, in
 						}
 					} else {
 #if TLS_SUPPORTED
-						if(check_tentative_tls(s->fd)) {
+						if((s->sat != TCP_CLIENT_DATA_SOCKET) && (s->sat != TCP_RELAY_DATA_SOCKET) && check_tentative_tls(s->fd)) {
 							s->tobeclosed = 1;
 							return -1;
 						}


Neustradamus:
@misi: Have you seen?

misi:
@tupelo-schneck 
This check AFAIU validates that if s->st is == TCP_SOCKET then it  should not be TLS, so we need to be careful because it could cause security issue. 
I am not original author of the code, so can you please help me and add a packet capture, or some detailed description how can I reproduce the issue, to be able understand it more.
Can you explain more why TCP_CLIENT_DATA_SOCKET and TCP_RELAY_DATA_SOCKET should not  checked by the check_tentative_tls(s->fd) function?

AFAIU there is a separate case to handle TLS connections
So I am not fully understand why the s->st is not set in your case to TLS_SOCKET in your use case. 

Can you please help me to understand it more. Thanks!!

tupelo-schneck:
This is all about [RFC6062](https://tools.ietf.org/html/rfc6062) TCP relays.

In our use case, a client wants to expose a TCP listener for a peer to connect to.  Forgive me for including the details below of how RFC6062 works, as I hope it might reveal the issue.

To start, the client initiates a "control connection" to coturn, and sends Allocate and then CreatePermission requests to instruct coturn to set up a TCP listener.  The client sends the address and port of coturn's listener to the peer, out of band.  

Now the peer opens a TCP connection to that listener.  Importantly, this socket will never carry TURN/STUN protocol traffic.  The peer need know nothing about TURN.  It's just opening a TCP connection to a listener.  The RFC calls this a "peer data connection"; I think in coturn, this is TCP_RELAY_DATA_SOCKET.

When coturn receives this connection, it goes back to the "control connection" to send a ConnectionAttempt indication to the client.  The client opens a brand-new TCP connection to coturn, and sends a ConnectionBind request over the new connection.  coturn now converts this new connection with the client into a "client data connection", TCP_CLIENT_DATA_SOCKET in coturn.

Quoting the RFC
> 4.5.  Sending and Receiving Data
> 
>    Once a client data connection is established, data sent on it by the
>    client will be relayed as-is to the peer by the server.  Similarly,
>    data sent by the peer to the server will be relayed as-is to the
>    client over the data connection.

(End of RFC6062 story.)

Now for our specific issue.  Imagine the client accepts HTTP traffic.  So it uses coturn to expose a listener; the peer now just sends totally standard HTTP requests to that listener, and (after the ConnectionBind request succeeds) the client receives totally standard HTTP requests on the client data connection, and just responds with totally standard HTTP responses.

That works perfectly!

But now, imagine that the client accepts *HTTPS* traffic.  Everything should work basically identically, with the peer sending a totally standard HTTPS request to the listener -- coturn should just relay the data as is, even though it happens to be TLS-wrapping-HTTP instead of HTTP.  Instead, we hit #605.

The solution is that coturn shouldn't care about the data on "client data connections" and "peer data connections" (TCP_CLIENT_DATA_SOCKET and TCP_RELAY_DATA_SOCKET) -- it should just relay the data as-is.  In particular it shouldn't reject them if they carry TLS data.

misi:
Thanks for the detailed explanation!

PR #655, opened 2020-11-28
Add support for proxy protocol V1

plinss:
I recently found that Nginx only supports Proxy Protocol V1 when initiating connections and Coturn only supports Proxy Protocol V2.

This patch adds support for Proxy Protocol V1. The code looks for Proxy Protocol V2 first, if that fails, then looks for V1 before aborting the connection.


misi:
Thanks for your help!

PR #659, opened 2020-12-09
fix 2 problems.

pig-peppa:
fix 2 problems.

### DIFF
--- a/src/client++/TurnMsgLib.h
+++ b/src/client++/TurnMsgLib.h
@@ -219,7 +219,7 @@ public:
 	 */
 	virtual ~StunAttr() {
 		if(_value)
-			free(_value,_sz);
+			free(_value);
 	}
 
 	/**
@@ -237,7 +237,7 @@ public:
 		if(sz>0xFFFF)
 			throw WrongStunAttrFormatException();
 		if(_value)
-			free(_value,_sz);
+			free(_value);
 		_sz = sz;
 		_value=(uint8_t*)malloc(_sz);
 		if(value)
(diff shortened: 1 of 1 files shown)


misi:
Removing exceptions may be not right?
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0003r5.html#4.0
See: Compatibility Concerns

src/client++/TurnMsgLib.h
```diff
@@ -75,7 +75,7 @@ class StunAttrIterator {
 	/**
 	 * Iterator constructor: creates iterator on raw messagebuffer.
 	 */
-	StunAttrIterator(uint8_t *buf, size_t sz) throw (WrongStunBufferFormatException) :
+	StunAttrIterator(uint8_t *buf, size_t sz) :
```
misi, line 78: Removing exceptions may be not right?
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0003r5.html#4.0
See: Compatibility Concerns

Missing try catch and throw exception..

hevc: exception classes should derive  from the std:exception
>     #include <stdexcept>
>     class MyException : public std::runtime_error {
>     public:
>       MyException() : std::runtime_error("MyException") { }
>     };
>     void f()
>     {
>        // ...
>        throw MyException();
>     }

pig-peppa: thank both of you, i'll fix it.

PR #661, opened 2020-12-10
fix 2 problems

### LINKED ISSUE #654: unsupport C++17
when i use libturnclient.a, compiler reports:
/usr/include/turn/TurnMsgLib.h:78:44: error: ISO C++1z does not allow dynamic exception specifications
  StunAttrIterator(u08bits *buf, size_t sz) throw (WrongStunBufferFormatException)

**can you fix it, when you get time.**

pig-peppa:
1>. fix #654.
2>. fix function 'void free(void*)' too many arguments.

### DIFF
--- a/src/client++/TurnMsgLib.h
+++ b/src/client++/TurnMsgLib.h
@@ -79,11 +79,11 @@ public:
 	StunAttrIterator(uint8_t *buf, size_t sz) :
 		_buf(buf), _sz(sz)  {
 		try {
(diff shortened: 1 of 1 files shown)


Neustradamus:
@misi: Can you look?

misi:
@pig-peppa Many thanks for contribution. I read docs and consulted with a more skilled C++ developer. 
I revert back to Your original idea, to just throw Errors and catch outside the lib. So I just removed dynamic exceptions.
Thanks for your patience..

PR #551, opened 2020-05-18
support of --acme-redirect <URL>

jelmd:
fix #546

PR #672, opened 2020-12-31
fix acme wrt. security, redundancy, consistency

jelmd:
The #551 PR rewrite/current implementation has several issues:
- unsafe copy of req:
  - the src pointer for the `req_url` copy may point to anything else, but not the intended location
  - the length of `req` is not checked and thus may overflow the `req_url` buffer
- too much redundant work
  - there is no need to search for space[s] or tab[s] to extract the URL path: `is_acme_req()` already takes care of it. It returns the position of the 2nd space of the request-line (see `plen`, RFC 2616, 5.1. The first space gets covered by the `GET_ACME_PREFIX`). If the request is not RFC conform, we do not care - ignore it.
  - since we already know the url (req + 4, plen), there is no need to copy those bytes around and no need for all the related helper vars
- inconsistency:
    - `GET_WELLKNOWN_ACMECHALLANGE_URL_PREFIX_LENGTH` got introduced in the rewrite. The corresponding string is missing. So a corresponding `#define` got introduced here.
    - If such a const gets defined, it should be used in _every_ related expression, otherwise it certainly causes more trouble than help.
    - awefully long names do not contribute to readability. So shortened it to an acceptable length, which does not hurt that much.
    - actually these defines are very specific implementation details and shouldn't exposed via `acme.h` or anything else.
- libevent fix:
  - at least on Ubuntu focal `int send_data_from_ioa_socket_nbh()` is unreliable and does not send the expected HTTP response for whatever reason (I guess, the "return socket" is already closed). So on ubuntu per default a workaround will be used to make sure, that a proper answer gets sent. To force `send_data_from_ioa_socket_nbh()` usage on ubuntu, one may  `export LIBEV_OK=1` before running `./configure ...`.
- license:
  Adding an outdated default header is IMHO not ok. Not sure, whether Citrix maintainers are still onboard, but I do not wanna get sued by anybody, just for re-using my own code. Therefore I added the MIT license header, which should be compatible with the project, does not require a change any other files, but still allows me (and others) to re-use it somewhere else.

### DIFF
--- a/configure
+++ b/configure
@@ -556,6 +556,17 @@ if [ "${SYSTEM}" = "NetBSD" ] ; then
 	fi
 fi
 
+# If acme_redirect does not work, send_data_from_ioa_socket_nbh() probably
+# does not work. Set LIBEV_OK=1 to use a workaround for it.
+if [ -z "${LIBEV_OK}" ]; then
+	LIBEV_OK=1
+	if [ "${SYSTEM}" = "Linux" ]; then
+		OS=$( lsb_release -si 2>/dev/null )
+		[ "${OS}" = "Ubuntu" ] || LIBEV_OK=0
+	fi
+fi
+[ "${LIBEV_OK}" = "1" ] && OSCFLAGS="${OSCFLAGS} -DLIBEV_OK"
+
 ###########################
 # Install shell commands
 ###########################
--- a/configure
+++ b/configure
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -x
+
 # Proprietary configure script of Coturn project
 
 cleanup() {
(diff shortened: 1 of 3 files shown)


misi:
Thanks for the corrections!
I did my best to merge and test your code. 
Sorry if I made more trouble and redundant thing it was due my limited knowledge, but I tested and seemed it was needed.
I tested it on Debian stable.

License: I used the common header, because I don't have info about this legal agreements, if it is still valid, or not.
Only the original author of the code @mom040267 can say more about it. 
IMHO one project with one license is more common and simple and understandable for the community. 
Fragmenting license per file is also uncommon, but I see your point if you want reuse your code anywhere else then I can go along with it. Do you really want to reuse this lines? Anyhow I would be more happy to not fragment, but it is just my 2 cents.

I do my best to serve this project, but I am very alone with my efforts, so I invite you to join me and help me maintain the project if you have time for it.

I will review and merge it soon. Sorry for the delay..

misi:
@jelmd If I am correct you mean -= here, right? see: https://github.com/coturn/coturn/commit/a28fee8cde3c115ad792909316ffd8e5256af2f1

jelmd:
Yes, good catch. :)

misi:
Thanks!!

PR #679, opened 2021-01-08
Readme.turnserver: how to run server as a daemon

rubo77:


README.turnserver
```diff
@@ -29,6 +29,13 @@ $ ./scripts/secure_relay.sh
 
 RUNNING THE TURN SERVER
 
+To run the coturn server as a daemon use:
+
+  $ turnserver -o
+
+Note that when running coturn as a daemon and you make any changes to the config file the server
+has to be reloaded.
```
wferi, line 37: If I understood correctly, there's no way to "reload" the daemon (more precisely: the configuration), so it must be **restarted** instead. And it doesn't matter whether coturn is running as a daemon or not.

misi: @wferi is right we should say restart not reload config..
Coturn only handles two signals:
* SIGUSR2 reload cert
* HUP reset log file

misi:
You also need to run `./make-man.sh` and `git add man` before `git commit` to update/regenerate the proper man pages too.

misi:
Many Thanks!

PR #684, opened 2021-01-13
Define OPENSSL_VERSION_1_1_1 on systems where it doesn't (yet) exist

brevilo:
Regression in 50ebef7

Preprocessor logic will fail and cause incompatible pointer type issues which effectively **breaks the turnserver** on all systems without `OPENSSL_VERSION_1_1_1` defined. Because the two processor logic statements fail semantically the wrong API/prototype gets used. That in turn causes [`turn_random()`](https://github.com/coturn/coturn/blob/master/src/apps/relay/ns_ioalib_engine_impl.c#L441) to choke (hang) on a thread mutex which means the server won't even start listening. Took me a while to figure this out.

The proposed solution just defines the *undefined* OpenSSL version in its lowest value such that the existing logic works as intended.


misi:
Thanks!

Issue #681, opened 2021-01-11
Question: What's the algorithm to hash the password in Redis?


wuelcas:
Hi everyone, a newbie to all WebRTC things here.

I have the task to configure our TURN server to use long-term credentials, for this we're creating a service that generates those credentials and using Redis to store the credentials. 

According to [the userdb schema for redis](https://github.com/coturn/coturn/blob/master/turndb/schema.userdb.redis) "For the long-term credentials there must be keys  "turn/realm/<realm-name>/user/<username>/key" and the values must be the the hmackeys", then it provides an example with a realm, user and a password but I can't figure out what hash algorithm should I use to generate the correct value to set in Redis, for the key it's pretty straight forward but I'm confused with the value. The doc says the values must be the hmackeys but I'm following what does it mean by that.

Thanks for the help in advance.

misi:
md5
`echo -n “user:realm:secret" | md5sum`

wuelcas:
Hey @misi thank you so much!

PR #687, opened 2021-01-14
Add hash algorithm for key value to redis userdb schema docs

wuelcas:
Hi.

A couple of days ago I asked what's the hash algorithm for the value stored in Redis for long-term credentials (#681). Thanks to @misi for the answer, it helped me a lot! but I think that's something that can be added to the docs in the schema.userdb.redis file. So that's pretty much what this PR does.


misi:
Thanks for your contribution!

PR #704, opened 2021-01-25
Packaging scripts can miss out on these errors

hills:
This script looks like it's designed to use a portable subset of
sh; I'm not aware of any portability problems of this patch.


misi:
Many Thanks! Sorry for the delay..

PR #710, opened 2021-01-29
debian/coturn.postrm: fix check for dpkg-statoverride on package removal

mika:
This fixes a regression introduced in commit b9a4a8.
When trying to remove a dpkg-statoverride, we need to check,
whether it exists. Otherwise it's failing, as can be observed
via piuparts runs:

| dpkg: unrecoverable fatal error, aborting:
|  unknown system group 'turnserver' in statoverride file; the system group got removed
| before the override, which is most probably a packaging bug, to recover you
| can remove the override manually with dpkg-statoverride
| E: Sub-process /usr/bin/dpkg returned an error code (2)

Closes: #981330

### DIFF
--- a/debian/coturn.postrm
+++ b/debian/coturn.postrm
@@ -6,6 +6,11 @@ TURNSERVER_USER=turnserver
 TURNSERVER_GROUP=turnserver
 
 if [ "$1" = "purge" ] ; then
+        TURNSERVER_CONFIG=/etc/turnserver.conf
+        if dpkg-statoverride --list "$TURNSERVER_CONFIG" >/dev/null 2>&1; then
+                dpkg-statoverride --remove "$TURNSERVER_CONFIG"
+        fi
+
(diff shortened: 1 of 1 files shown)


wferi:
Good catch! I think the statoverride removal should move before the account removal as well.

mika:
> Good catch! I think the statoverride removal should move before the account removal as well.

Took care of it, new PR version just pushed :)

mika:
Friendly ping, we should ensure this fix makes it into Debian/bullseye :)

mika:
FYI, if you want to get the package included in bullseye without going for an exception via the release team, please consider uploading it *now* :)

wferi:
It isn't that urgent, the incoming soft freeze won't block automatic migrations, just slow them down.
Don't worry, we'll take care of this.

misi:
Thanks @mika !
We will try to make it.

PR #718, opened 2021-02-24
mongoc version 1.17.4

dcharbonnier:
**The current version of mongoc is old, many options of the connection string are not available.**

This pr change the way the image is built:
- use debian slim (this and other optimizations reduce the production image size, 750Mo =>  130Mo)
- add a stage production (I don't see the point of installing mongodb inside the image, maybe something I didn't get)
- build mongoc to get a 1.17.4 version
- remove the exposed ports on the production image, if you use a network host you don't need them and this makes the start of the image really slow even if the user doesn't need all those ports exposed.

### DIFF
--- a/docker/coturn/Dockerfile
+++ b/docker/coturn/Dockerfile
@@ -1,12 +1,36 @@
-### 1. stage: create build image
-FROM debian:stable AS coturn-build
+### 1. stage: create mongoc image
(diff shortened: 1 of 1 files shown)


misi:
Thanks!

misi:
Sorry for the delay..

PR #719, opened 2021-02-26
Use turn_mutex instead of pthread_mutex_t

KangLin:
Be will support windows


KangLin:
Use turn_mutex instead of pthread_mutex_t, in order to be will compatible windows.

KangLin:
This turn_mutex_init_recursive is not available in MSVC.
 Is it really needed in this project?
 If it is not necessary, consider removing ？

KangLin:
Merge to #855

PR #727, opened 2021-03-04
Support older mysql client version in configure

JoKoT3:
Hi,

on a system with an older version of mysql and mysql-libs (RHEL6 / mysql version 5.1.x) there is no pkg-config file available for mysql.

This patch fallback on using mysql_config if it is found in PATH.


misi:
Thanks!

PR #739, opened 2021-03-23
SSL reload has hidden bugs which cause crashes

hills:
The process  would run for a while, but eventually crash. valgrind shows a bug immediately, but memory (re-)use means that actual effect can be mostly invisible (tested on musl libc).

Each patch explains. We've been running these patches for a while now, but I just rebased them to the most recent coturn.

### DIFF
--- a/src/apps/relay/dtls_listener.c
+++ b/src/apps/relay/dtls_listener.c
@@ -935,36 +935,6 @@ static int init_server(dtls_listener_relay_server_type* server,
   server->verbose=verbose;
   
   server->e = e;
-  
-#if DTLS_SUPPORTED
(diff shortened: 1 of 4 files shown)


misi:
Many Thanks! I need some time to understand the changes! Thanks for your patience..

argggh:
As chance would have it, we have also been seeing this in the wild in the last 24 hours. FWIW, the change looks reasonable to me. (I was probably  the one who introduced this as well way back when... :disappointed:)

hills:
Any movement on these patches? I would not want them to get out of date by some change on the main branch.

misi:
@hills I will merge it ASAP..

PR #753, opened 2021-04-19
Use DNS requests to discover external IP address

j1elo:
Using DNS requests is a much more robust and reliable method to discover a machine's external IP address, instead of the previous method of using Curl against an HTTP service.

Using Curl is fine, but the kind of services that are typically used (here it was icanhazip.com, but there are lots more with a similar behavior) are are not as dependable as the official DNS request methods supported by some of the biggest service providers.

This uses `dig`, which Alpine provides in the package "bind-tools" and Debian in "dnstools".

The script discovers IPv4 address by default, but can also be made to look for IPv6 addresses (with `--ipv6`), because I'm just adapting the code directly from a script that I already had been using for some time (https://github.com/j1elo/shell-snippets/blob/master/discover-public-ip.sh). But if IPv6 is not desired, that part of the code could be easily removed, just let me know.


j1elo:
I've adapted my PR from here: https://github.com/instrumentisto/coturn-docker-image/pull/41
which happened just when the migration from instrumentisto to this project was under way.

Before it was just for Alpine, but now that Debian is also an option for Docker images, I've adapted the script so it works the same in both: Alpine uses `ash`, while Debian uses `dash`, and actually ash supports more options for strict error checking, so I ended up making a generic method to enable as many as possible on each shell.

j1elo:
> *Some checks were not successful*
>  Docker CI / buildx (debian, true, false) (pull_request) Failing after 111m — buildx (debian, true, false)

Are these test failures caused by the changes in the PR? I cannot see any reference to "`external-ip`" in the logs.

tyranron:
@j1elo it seems that Docker layers cache on CI is not doing well. Let's try to opt-out it.

tyranron (changes requested):
@j1elo thanks once again for such nice contribution!

I've altered sources a bit to avoid script duplication.

It would be also nice:
- to mention IPv6 capability in README
- to add some simple tests checking the `detect-external-ip` is presents and runs OK

tyranron:
Ooopps... that's an accident!

tyranron:
@j1elo now I cannot restore the git log, as GitHub doesn't allow me to push to your `master` brach anymore (as it's not bound to this PR). Please, push any commit to your `master` branch, so I can reopen this MR and restore original commits.

j1elo:
I'm a bit confused, should I push anything at all?
I literally did an [empty commit](https://stackoverflow.com/questions/20138640/pushing-empty-commits-to-remote/61789697#61789697) right now, hope that is enough?

tyranron:
@j1elo yup, that did the trick. Thanks! The PR now is fully restored and we can continue our work on it.

j1elo:
Nice! also I agree, it's better to not have the script duplicated, so having it only once is already an improvement in itself :)

Github tells me that there are changes requested, but when I click on "View changes", they complain that "couldn’t find those commits"... maybe because of the closing and reopening of the PR.

Is there anything else that should be added? Maybe the tests you mentioned. Was the previous script being tested? if not, maybe this PR should be kept small as in only changing things that already existed, and add tests as a new, different change?

Otherwise to add tests I'm not sure how to go about that. The Docker image would need to be built inside a NAT'ed machine, of which the real external IP is well known, and then compare it to the result of the script (either this one or the old one, doesn't matter)

tyranron:
@j1elo 

> Was the previous script being tested? if not, maybe this PR should be kept small as in only changing things that already existed, and add tests as a new, different change?

No, there were no previous tests for this feature, and it's a good opportunity to finally land them  🙃

PR is small enough, so covering with tests won't make it too large. And as the tests are relevent to this specific feature, they don't deserve a separate PR, in my opinion.

> Otherwise to add tests I'm not sure how to go about that. The Docker image would need to be built inside a NAT'ed machine, of which the real external IP is well known, and then compare it to the result of the script (either this one or the old one, doesn't matter).

Well, I think we don't need to test whether the IP is resolved correctly here. The aim is rather to avoid future accidental regressions in the image and this script. So, to check whether the script is present in the image, and to check whether it runs OK (finishes with `0` exit code) would be quite enough. Additionally, it would be nice to check that the returned result is IPv4/IPv6-like (by mask) without checking the exact equality of the returned value.

------------------------------

And mentioning IPv6 capability for `detect-external-ip` command in README would be nice as well.

Thanks!

tyranron (approved):
@misi I've added the tests and mention. Please, review.

tyranron:
ping @misi

tyranron:
@misi I've decided to merge it to produce a new Docker image tag. If you'll find some issues with this PR, I'l resolve them in a subsequent one.

Issue #730, opened 2021-03-08
CoTurn stopped functioning suddenly


Chandramouli-P:
Hello,

CoTurn (v.4.5.2), Ubuntu 18.04 server, AWS. 

I am using CoTurn in 3 servers. I am clearly noticing that CoTurn is stopped functioning frequently. Please note that CoTurn service is running and not seeing any errors in the log file. When I restarted the server, it is starting functioning again. 

Can anyone tell what could be the root cause and suggest a solution? Thank you.

Best Regards,
Chandramouli.

Chandramouli-P:
Hello,

Good morning and any help would be appreciated. Thank you.

Best Regards,
Chandramouli.

Chandramouli-P:
Hello Team,

Good morning and any help would be appreciated. Thank you.

Best Regards,
Chandramouli.

Chandramouli-P:
Hello Team,

Good morning and any help would be appreciated. Thank you.

Best Regards,
Chandramouli.

hakicode:
I am also having the same problem on the latest automatic installation on ubuntu, I need your help as well, hope someone can help us.

Chandramouli-P:
Hello Team,

Good morning and any help would be appreciated. Thank you.

Best Regards,
Chandramouli.

Chandramouli-P:
Hello Team,

Good morning and any help would be appreciated. Thank you.

Best Regards,
Chandramouli.

Mart-Bogdan:
Hello, @Chandramouli-P. I have no idea how coturn work, but I could suggest some workaround.

If it's working for few days, and restart helps. I'd suggest creating crontab job that would restart daemon (not the whole server) once a day. Dirty workaround, but that should do the work. I've used a similar hack in production. Perfectly would be to set up restarts on not working hours.

Chandramouli-P:
Hello @Mart-Bogdan ,

Good morning and thank you for your suggestion. It seems that this issue is resolved. I am not seeing this behavior from last couple of days. If it repeats again, I will follow your suggestion. Thank you.

Best Regards,
Chandramouli.

PR #755, opened 2021-04-26
Fix sqlite shutdown race

moznuy:
Remove sqlite3_shutdown() 

Because:
1. According to sqlite3 docs sqlite3_initialize() and sqlite3_shutdown() do not have to be invoked
2. sqlite3_initialize() is never called explicilty
3. sqlite3_shutdown() is not threadsafe and sqlite_init_multithreaded is not called holding a lock
4. According to docs all connections must be closed before invoking sqlite3_shutdown() but they are not (from the different threads).

Basically my team were experiencing sudden loss of service after a while.
I started to look into it. On local machine I was getting:
1. SEGFAULT occasionally. (segfaults in sqlite lib)
2. loss of service
3. normal work

These changes removed for me first two points.

Maybe related to #730


misi:
Many thanks! Sorry for my late reply!

misi:
Many thanks for your PR and changes it is merged...

PR #771, opened 2021-05-21
Performance: Don't bzero the entire stun_buffer_list_elem

haseebq:
On high traffic servers bzero -> memset takes a significant amount of time because each stun_buffer_list_elem can be ~64kb


haseebq:
image
image

In the above flamegraphs you can that the majority of the time udp_server_input_handler spends is in ioa_network_buffer_allocate -> __memset_avx2_erms

haseebq:
Looks like the reason this is done is because of an information disclosure CVE due to length padding. This patch will cause a regression. Closing pull request

PR #773, opened 2021-05-25
Fix for performance regression caused by CVE-2020-4067 fix

haseebq:
Overall flamegraph:image

Zoom in on udp_server_input_handler:
image

Zoom in on socket_input_handler:
image

In the above flamegraphs you can see that the majority of time in **udp_server_input_handler** and **socket_input_handler** is being spent in memset. This is because both eventually call **new_blist_elem()** which ends up zeroing the entire stun_buffer_list_elem struct before returning. The stun_buffer_list_elem is almost 64kb in size and this causes performance issues on high traffic servers.

With this change, we zero out any padding bytes explicitly in stun_attr_add_str() and avoid zeroing out the entire stun buffer.

### DIFF
--- a/src/apps/relay/ns_ioalib_engine_impl.c
+++ b/src/apps/relay/ns_ioalib_engine_impl.c
@@ -297,15 +297,15 @@ static stun_buffer_list_elem *new_blist_elem(ioa_engine_handle e)
 
 	if(!ret) {
 	  ret = (stun_buffer_list_elem *)malloc(sizeof(stun_buffer_list_elem));
-	  if (ret) {
-		ret->next = NULL;
-	  } else {
-		TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "%s: Cannot allocate memory for STUN buffer!\n", __FUNCTION__);
-	  }
 	}
 
 	if(ret) {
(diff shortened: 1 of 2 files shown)


misi:
Many Thanks!

misi:
@wferi what is your opinion? I need to be careful here because I am not totally sure that on architectures where the memory alignment is different. It does not leak, and I want to avoid adding back a leak if some alignment or padding not clear memory.

Neustradamus:
@misi, @wferi Any news?

andresbersier:
@misi, @wferi hi, are there any plans to merge this? I've seen a 20% difference in CPU usage in certain scenarios with/without this fix.

ggarber:
The change looks safe to me in terms of memory alignment and looks like a reasonable improvement.

My only concern is if there are other places in the coturn codebase where the buffer is returned to the users without initialization apart from the one being fixed here in `stun_attr_add_str`.    But it looks like nobody from the three people commenting this ticket had issues with that and the performance improvement is big so I'm thinking on landing it and see if there is any problem identified by anybody while using it for the next weeks.

Thank you for the work here @haseebq

eakraly:
Wonder if @haseebq can create another flame graph after his fix. Great improvement!

Issue #786, opened 2021-06-28
user quota release 


yfaker:
when user quota check is enabled, user's quota info is recorded in map(rp->status.alloc_counters) according to check_new_allocation_quota function, and seems to be never deleted from the map.
as the map stucture is LIST, the map will become longer and longer when new incoming users continues, and the user quota check processing will takes more time

yfaker:
in release_allocation_quota function, when user's quota count becomes 0, should we delete the record like this? or maybe I have some misunderstanding?
'
if (value) {
    value = (ur_string_map_value_type)(((size_t)value) - 1);
    if (value == 0) {
          ur_string_map_del(rp->status.alloc_counters, (ur_string_map_key_type) username);
      } else {
          ur_string_map_put(rp->status.alloc_counters, (ur_string_map_key_type) username, value);
      }
}
'

eakraly:
Fixed in #792

PR #792, opened 2021-07-15
Fix user quota release #786

yfaker:


Neustradamus:
@misi: Happy New Year 2022!

Can you look the @yfaker PR?

Thanks in advance.

Neustradamus:
@ggarber: Can you look?

eakraly:
Following this as well
This is also a memory leak if I read this correctly

ggarber:
Yep, Looks like a bug but I'm not sure of the side effects of this change yet.    Are we sure nobody will read from that map after that point in the code?  Is this the only/best place where we need to do the `ur_string_map_del` ?

I will try to take a look at those things asap.  If @yfaker have any input on those things that will be also helpful.

PR #812, opened 2021-08-22
Don't set socket buffer size for UDP socket

mmasaki:
`UR_CLIENT_SOCK_BUF_SIZE` and `UR_SERVER_SOCK_BUF_SIZE` are too small for high bandwidth (approximately 4Mbps or higher).
It causes UDP packet receive errors.
This change removes `set_sock_buf_size()` from `set_socket_options_fd()`.
The socket buffer size is determined by the kernel and it is controllable through the sysctl command etc.

### DIFF
--- a/src/apps/peer/udpserver.c
+++ b/src/apps/peer/udpserver.c
@@ -84,8 +84,6 @@ static int udp_create_server_socket(server_type* server,
     TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO,"Cannot bind udp server socket to device %s\n",server->ifname);
   }
 
-  set_sock_buf_size(udp_fd,UR_SERVER_SOCK_BUF_SIZE);
-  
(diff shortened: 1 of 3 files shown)


eakraly:
UR_SERVER_SOCK_BUF_SIZE is 32*65535 which is 2MB - that is pretty big buffer

Even so, in some environments dealing with kernel buffers is difficult and unclear so being able to control it for coturn makes sense.
I would have made into configuration variable so that it can be set through CLI or config file (and 0 to turn it off and use default kernel config)

ggarber:
@mmasaki Any chance you could make it configurable as proposed by @eakraly ?

eakraly:
As an alternative, we can open an issue instead  - to make it configurable

ggarber:
I'm working on this in this branch: https://github.com/coturn/coturn/pull/1089

@mmasaki Will close this PR for now if that's ok for you but contributions as code, reviews or testing are more than welcomed.

mmasaki:
@ggarber  Closing. Thank you!

PR #825, opened 2021-09-20
dbd_sqlite: Don't shutdown sqlite3 db during init

argggh:
Ref https://www.sqlite.org/c3ref/initialize.html:

    The sqlite3_initialize() interface is threadsafe, but
    sqlite3_shutdown() is not.

We currently call `sqlite3_shutdown` from all threads as part of `sqlite_init_multithreaded`, and this has been observed to have adversarial effects during startup if many threads receive their first inbound request at the same time. The apparent motivation behind calling shutdown is to make the subsequent calls to `sqlite3_config` succeed, since these ordinarily return SQLITE_MISUSE if called multiple times. However, this function is also documented to not be thread safe, so introduce a barrier that ensures we only initialize once over all threads.

### DIFF
--- a/src/apps/relay/dbdrivers/dbd_sqlite.c
+++ b/src/apps/relay/dbdrivers/dbd_sqlite.c
@@ -45,11 +45,13 @@
 //////////////////////////////////////////////////
 
 static pthread_mutex_t rc_mutex = PTHREAD_MUTEX_INITIALIZER;
+static pthread_mutex_t init_mutex = PTHREAD_MUTEX_INITIALIZER;
 static pthread_cond_t rc_cond = PTHREAD_COND_INITIALIZER;
 
 static int read_threads = 0;
 static int write_level = 0;
 static pthread_t write_thread = 0;
+static int sqlite_initialized = 0;
 
 static void sqlite_lock(int write)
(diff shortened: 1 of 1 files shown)


misi (commented):
I am wondering if non thread safe sqlite3_shutdown was removed, then do we still need the new mutex?
In the new mutex all actual functions calls seems to me thread safe. May I miss something?

argggh:
The mutex ensures that only one thread runs `sqlite3_config`, which is also not threadsafe. We could probably release it somewhat earlier than we do here, but this is only run once in each thread so the performance impact should be negligible.

misi:
Thanks and sorry you are right.

misi:
I realized and alternative fix in PR#755 so it seems more simple so I reverted your change in favor solution in #755. 
Let me know if you agree that the two fix is identical, and it is Ok for you to apply that fix instead of yours.

argggh:
Yes, that looks like a conceptually simpler approach. I haven't tested it in the field, but offhand it is hard to see why it shouldn't be equivalent.

argggh:
(I wish I'd seen that PR earlier, mea culpa.)

PR #849, opened 2021-11-09
Fix build with libressl 3.4.0+

rozhuk-im:
SSL will become opaque in LibreSSL 3.4.x, hence the code reaching inside
it will result in build breakage.
SSL_set0_rbio() now available in libressl.


Neustradamus:
@misi: Happy New Year 2022!

Can you look the @rozhuk-im PR?

Thanks in advance.

Neustradamus:
@ggarber: Can you look here?

ggarber:
The change looks correct and I don't know how to verify it so I'm landing it and reverting if somebody detects any problem.

Thx a lot for the fix @rozhuk-im !

PR #865, opened 2021-12-14
Store git version information in Docker image metadata

davel:
I have added labels to the image so that it is explicit which revision was used to build a container.

I have also tweaked the Dockerfile to work better when you have forked the repository.


tyranron (approved):
@davel thanks!

PR #867, opened 2021-12-16
Add port 5349 to Docker readme

GingerAdonis:
The default Docker coturn config enables usage of port 5349. I'd say the readme needs to reflect that.


tyranron (approved):
@GingerAdonis thanks!

PR #870, opened 2021-12-26
configure: support MariaDB Connector/C

micmac1:
If libmariadb is installed from the MariaDB server package, the pc file
is "mariadb.pc". But when MariaDB Connector/C is used, it's actually
"libmariadb.pc". This commit adds the latter to the detection.

Signed-off-by: Sebastian Kemper <sebastian_ml@gmx.net>

### DIFF
--- a/configure
+++ b/configure
@@ -945,7 +945,7 @@ fi
 ###########################
 
 if [ -z "${TURN_NO_MYSQL}" ] ; then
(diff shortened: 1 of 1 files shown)


Neustradamus:
@misi: Happy New Year 2022!

Can you look the @micmac1 PR?

Thanks in advance.

ggarber (approved):

PR #887, opened 2022-02-10
Update hiredis_libevent2.c

oasangqi:
- fix redis read message queue bug when reconnected to redis server;


ggarber:
@oasangqi thx for the fix.  Can you add to the description what is the bug, how this change fixes it and how you tested it?  Thank you very much!

ggarber:
@oasangqi any update? Otherwise i think i will close the PR for now and will reopen when more info is provided if that's ok.    
I took a look at hiredis doc and can't figure out the exact impact of this flag and why is it needed in one of the events and not the others.

Issue #666, opened 2020-12-16
Running on docker with prometheus open keep eating up memory


yuanchieh-cheng:
It seems that prometheus keeps all relay connection records in memory, so the CPU and memory keep increasing.  
After reboot the server, CPU / memory drop down.  

Is there any way to clear the memory without restarting the server?
It's ok to clear all the records.

domeger:
@yuanchieh-cheng how did you enable the endpoint ?, via the config ?

yuanchieh-cheng:
via config

misi:
@yuanchieh-cheng Can you please find the root and fix it and send a patch Pull Request. 
I work on other issues, don't find the time to check it in short term...

yuanchieh-cheng:
I am not the c programmer but I would try my best.   
My propose is to flush all the data if the api /metrics is called successfully.  
Is it ok to do so ?

misi:
@wolmi  Can you please help on the fix of the issue? 
See: https://github.com/coturn/coturn/pull/517

misi:
@yuanchieh-cheng I don't have a prometheus setup, so I don't fully see yet what data's stuck in memory. 
Can you dig deeper in a memory dump and give more info about what info's/data block are stuck in memory and eating the memory?

yuanchieh-cheng:
Ok, I would try to produce the debug data

wolmi:
I think it is related to the huge amount of allocations and the fact that the allocation is a label that causes to have a lot of information on each metric.

The use case is a test environment or is it production? can you share some info about the number of users?

yuanchieh-cheng:
Yes I think that is the root cause of the issue.
I still run in the test but I think it should be environment unrelated issue.  
I run a cronjob at 1 min rate to start TCP relay for health check.  
Roughly after one day (86400 records generated), it ate up 200 MB roughly.  
Even worst I found the container throw Segment fault(core dump).

deyceg:
> I think it is related to the huge amount of allocations and the fact that the allocation is a label that causes to have a lot of information on each metric.
> 
> The use case is a test environment or is it production? can you share some info about the number of users?

@wolmi We came to the same conclusion. Allocation ID is a very bad, high cardinality label. It should be removed otherwise the exporter is not usable.

wolmi:
[...]

Issue #827, opened 2021-09-29
When TURN REST API used Prometheus metrics send wrong user name


udovichenkoAlexander:
When TURN REST API used then Prometheus metrics user name send as **TIME:USER**. 
As result users not grouped correctly in monitoring and this lead to errors like **max-values-per-tag limit exceeded** in influx. 
It looks like user part must be extracted in this case before send metric username

![Снимок экрана от 2021-09-29 13-14-24](https://user-images.githubusercontent.com/23613015/135249824-aed19ade-0c2c-4c8e-b0f7-1688d2bf1f58.png)

misi:
This is the right thing to log. username=timestamp:appdata 
see https://www.ietf.org/proceedings/87/slides/slides-87-behave-10.pdf

udovichenkoAlexander:
Yes, I understand how REST API AUTH works.
Ok, it seems that to properly aggregate statistics by users and not by user tokens, I need truncate EXPIRY DATE by myself

misi:
Sorry if you have any better proposal how to make it differently, we are open to improve the actual behavior.
Please make a PR and we can make an option to only put appdata in the reports.

udovichenkoAlexander:
Ok, If I choose this solution I will make a PR

PR #919, opened 2022-06-20
Configurable username labelling in Prometheus metrics

sysvinit:
When Prometheus metrics are enabled, traffic metrics are labelled with client username, which permits tracking traffic usage on a per-client basis. However, the TURN REST API authentication scheme uses ephemeral usernames, and so authenticating with a new username will cause a new label value to be generated. As prometheus-client-c doesn't garbage collect old metrics labels, this means that memory usage will grow over time, corresponding to clients authenticating with new usernames.

This change adds a new option `--prometheus-no-username-labels`, which will disable the use of username labels when traffic metrics are record, which should eliminate the memory leak issue. This should address issue #666.

### DIFF
--- a/README.turnserver
+++ b/README.turnserver
@@ -284,9 +284,9 @@ Flags:
  --prometheus		Enable prometheus metrics. By default it is
 			disabled. Would listen on port 9641 unther the path /metrics
 			also the path / on this port can be used as a health check
- --prometheus-no-username-labels	Disable labeling prometheus traffic
+ --prometheus-username-labels	Enable labeling prometheus traffic
 			metrics with client usernames. Labeling with client usernames is
-			enabled by default, however this may cause memory leaks when using
+			disabled by default, beacuse this may cause memory leaks when using
 			authentication with ephemeral usernames (e.g. TURN REST API).
 
 -h			Help.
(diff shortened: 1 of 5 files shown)


ggarber:
Shouldn't the default be no-username-labels?   It looks like in 99% of the cases it would be a bad idea to have thousands or millions of labels.

sysvinit:
> Shouldn't the default be no-username-labels? It looks like in 99% of the cases it would be a bad idea to have thousands or millions of labels.

I wanted to preserve the existing default behaviour as a principle-of-least-astonishment thing, which is why I made the username labelling opt-out. However, if you think that it's worth changing the defaults here I can invert the logic so that the labels are opt-in.

ggarber:
> > Shouldn't the default be no-username-labels? It looks like in 99% of the cases it would be a bad idea to have thousands or millions of labels.
> 
> I wanted to preserve the existing default behaviour as a principle-of-least-astonishment thing, which is why I made the username labelling opt-out. However, if you think that it's worth changing the defaults here I can invert the logic so that the labels are opt-in.

Yes, please.  I think it is better to make that change to limit the amount of people using the default settings and having memory issues as soon as they start having real traffic in production.

If you can invert it I'll try to review it tomorrow and push it.  Other than that the PR looks perfect.  Thank you!

sysvinit:
See 6fd08ba, as requested :)

ggarber:
Looks great @sysvinit.  I'll try to do some testing as soon as possible and merge it.  Thank you!

udovichenkoAlexander:
also relaterd with https://github.com/coturn/coturn/issues/827

PR #942, opened 2022-08-15
Disable SSL renegotiation

### LINKED ISSUE #892: SSL/TLS: Renegotiation DoS Vulnerability (CVE-2011-1473, CVE-2011-5094)
Running coturn 4.5.2 (and openssl 1.1.1m) our latest scan with the Greenbone Source Edition (Vulnerability Management solution based/forked from OpenVAS) is reporting the issue mentioned below for port `3478/tcp` & `443/tcp` on which coturn listens so the question rises if this is a valid security issue or just a false positive?

turnserver.conf is configured like:

```
listening-ip=0.0.0.0
listening-port=3478
tls-listening-port=443
fingerprint
use-auth-secret
static-auth-secret=xxx
realm=domain.com
total-quota=100
bps-capacity=0
stale-nonce=600
cert=/etc/ssl/cert.crt
pkey=/etc/ssl/key.key
cipher-list="EECDH+ECDSA+AESGCM EECDH+aRSA+AESGCM EECDH+ECDSA+SHA384 EECDH+ECDSA+SHA256 EECDH+aRSA+SHA384 EECDH+aRSA+SHA256 EECDH+aRSA+RC4 EECDH EDH+aRSA !aNULL !eNULL !LOW !3DES !EXP !PSK !SRP !DSS !RC4"
simple-log
no-cli
no-multicast-peers
no-tlsv1
no-tlsv1_1
```

### Summary

The remote SSL/TLS service is prone to a denial of service (DoS) vulnerability.
The following indicates that the remote SSL/TLS service is affected:

### Detection result

```
Protocol Version | Successful re-done SSL/TLS handshakes (Renegotiation) over an existing / already established SSL/TLS connection
----------------------------------------------------------------------------------------------------------------------------------
TLSv1.2          | 10
```

### Details
The flaw exists because the remote SSL/TLS service does not
properly restrict client-initiated renegotiation within the SSL and TLS protocols.

Note: The referenced CVEs are affecting OpenSSL and Mozilla Network Security Services (NSS) but
both are in a DISPUTED state with the following rationale:

> It can also be argued that it is the responsibility of server deployments, not a security
library, to prevent or limit renegotiation when it is inappropriate within a specific environment.

Both CVEs are still kept in this VT as a reference to the origin of this flaw.

### Detection method
Checks if the remote service allows to re-do the same SSL/TLS handshake (Renegotiation) over an existing / already established SSL/TLS connection.
Details: SSL/TLS: Renegotiation DoS Vulnerability (CVE-2011-1473, CVE-2011-5094...OID: 1.3.6.1.4.1.25623.1.0.117761)
Used version: 2021-11-15T10:28:20Z

Affected Software/OS:
Every SSL/TLS service which does not properly restrict client-initiated renegotiation.

Impact:
The flaw might make it easier for remote attackers to cause a DoS (CPU consumption) by performing many renegotiations within a single connection.

Solution:
Users should contact their vendors for specific patch information.

A general solution is to remove/disable renegotiation capabilities altogether from/in the affected
SSL/TLS service.

Sources: CVE-2011-1473, CVE-2011-5094

eakraly:
Disable SSL renegotiation by setting `SSL_OP_NO_RENEGOTIATION` on SSL connection (both TLS and DTLS)

Fixes #892

Tested by connecting to running server and asking for renegotiation:
```
#openssl s_client -connect localhost:443
...
---

RENEGOTIATING
4398376448:error:14094153:SSL routines:ssl3_read_bytes:no renegotiation:ssl/record/rec_layer_s3.c:1561:
```
and DTLS
```
#openssl s_client -connect localhost:443 -dtls1
...
---

RENEGOTIATING
4572587520:error:14102199:SSL routines:dtls1_read_bytes:too many warn alerts:ssl/record/rec_layer_d1.c:586:
```

### DIFF
--- a/src/apps/relay/dtls_listener.c
+++ b/src/apps/relay/dtls_listener.c
@@ -295,7 +295,7 @@ static ioa_socket_handle dtls_server_input_handler(dtls_listener_relay_server_ty
 	SSL_set_accept_state(connecting_ssl);
 
 	SSL_set_bio(connecting_ssl, NULL, wbio);
-	SSL_set_options(connecting_ssl, SSL_OP_COOKIE_EXCHANGE);
+	SSL_set_options(connecting_ssl, SSL_OP_COOKIE_EXCHANGE | SSL_OP_NO_RENEGOTIATION);
 
 	SSL_set_max_cert_list(connecting_ssl, 655350);
 
@@ -581,7 +581,8 @@ static int create_new_connected_udp_socket(
 
 		SSL_set_bio(connecting_ssl, NULL, wbio);
 
-		SSL_set_options(connecting_ssl, SSL_OP_COOKIE_EXCHANGE);
+		SSL_set_options(connecting_ssl, SSL_OP_COOKIE_EXCHANGE | SSL_OP_NO_RENEGOTIATION);
+
 		SSL_set_max_cert_list(connecting_ssl, 655350);
 		int rc = ssl_read(ret->fd, connecting_ssl, server->sm.m.sm.nd.nbh,
 				server->verbose);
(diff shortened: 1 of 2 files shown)


ggarber (approved):
Awesome!

PR #943, opened 2022-08-16
[turncommon] depend on OpenSSL

### LINKED ISSUE #242: 'openssl/md5.h' file not found
OS: Mac 10.13.3
Description: Hi.I had openssl installed by ``brew`` command, and after i run make in the coturn directory, there is an error emerged below:
```
src/apps/common/ns_turn_openssl.h:34:10: fatal error: 'openssl/md5.h' file not found
#include <openssl/md5.h>
         ^~~~~~~~~~~~~~~
1 error generated.

```

jeremy-murphy:
Maybe only one is strictly required, but the list of headers used in `ns_turn_openssl.h` includes things from both targets.

I think this correctly resolves #242. I can now compile coturn on MacOS without having to manually specify any `-I` include directory for OpenSSL. (I do specify `OPENSSL_ROOT_DIR`, of course.)

How come this works on Linux? Is the `Libevent::openssl` target forwarding the OpenSSL include directories?
That's another line of investigation if you don't like this solution.

### DIFF
--- a/src/apps/common/CMakeLists.txt
+++ b/src/apps/common/CMakeLists.txt
@@ -37,6 +37,11 @@ if(hiredis_FOUND)
 else()
     list(APPEND COMMON_DEFINED TURN_NO_HIREDIS)
 endif()
+
+find_package(OpenSSL REQUIRED Crypto SSL)
+list(APPEND COMMON_LIBS OpenSSL::Crypto)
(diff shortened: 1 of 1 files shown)


ggarber:
It makes sense to make, thank you very much @jeremy-murphy

PR #944, opened 2022-08-17
Create github action to build PR using cmake

eakraly:


.github/workflows/cmake.yaml
```diff
@@ -0,0 +1,35 @@
...
+            libssl-dev \
+            libpq-dev libmariadb-dev libsqlite3-dev \
+            libhiredis-dev \
+            libmongoc-dev \
+            libmicrohttpd-dev
+    - uses: actions/checkout@v3
+    - name: Configure CMake
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}}
+    - name: Build
+      run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}}
```
ggarber, line 32: Do we need to pass the BUILD_TYPE both to cmake and make?  I thought it was required only in one place.

ggarber (approved):

.github/workflows/cmake.yaml
eakraly: This was copied from github actions sample. I agree there is duplication and it looks unnecessary

PR #949, opened 2022-08-20
Fix cmake build on macOS

### LINKED ISSUE #946: Building coturn on macOS
Now that there are 2 mechanisms to build coturn (configure and cmake) I have following:
- configure works perfectly (provided I pass --prefix=/usr/local/opt/openssl@1.1 where openssl@1.1 is from brew)
- CMake is struggling with all the dependencies (openssl, hires, mongo-c, etc), but since only openssl is required then `-DOPENSSL_ROOT_DIR=/usr/local/opt/openssl@1.1/` fixes it
- CMake build does not pass because of missing `pthread_barrier_t` `pthread_barrier_wait` `pthread_barrier_init` - what is the solution to that?

@jeremy-murphy you submitted a great improvement to CMake build lately so you can probably shed some more light on building coturn using CMake on macOS?

eakraly:
macOS does not have `pthread_barrier_*` so need to define `TURN_NO_THREAD_BARRIERS` as a workaround

Fixes #946 

Tests plan: 
- run cmake to generate make files
- run make to build turnserver
- run `examples/run_tests.sh` and pass successfully

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,8 +56,8 @@ IF(MSVC)
 ENDIF(MSVC)
 
 IF(APPLE)
-	add_compile_definitions("TURN_NO_THREAD_BARRIERS=1")
-endif()
+    add_compile_definitions("TURN_NO_THREAD_BARRIERS=1")
+ENDIF()
 
 include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,10 @@ IF(MSVC)
     add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
 ENDIF(MSVC)
 
+IF(APPLE)
+	add_compile_definitions("TURN_NO_THREAD_BARRIERS=1")
+endif()
+
 include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
 include(GenerateExportHeader)


eakraly:
@jeremy-murphy FYI
Thanks you for the idea on how to make it work!

CMakeLists.txt
```diff
@@ -55,6 +55,10 @@ IF(MSVC)
     add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
 ENDIF(MSVC)
 
+IF(APPLE)
+	add_compile_definitions("TURN_NO_THREAD_BARRIERS=1")
+endif()
```
ggarber, line 60: consistent casing plz "IF/ENDIF"

ggarber (approved):

CMakeLists.txt
```diff
@@ -55,6 +55,10 @@ IF(MSVC)
     add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
 ENDIF(MSVC)
 
+IF(APPLE)
+	add_compile_definitions("TURN_NO_THREAD_BARRIERS=1")
```
ggarber, line 59: Is the indentation the same as in rest of the file (looks like tabs vs spaces but maybe it is just the GH UI).

CMakeLists.txt
eakraly: Right, I had reservations about that
I see different casing in different CMakeList files.

Issue #950, opened 2022-08-21
memory leak if not define  TURN_NO_TLS


huhaipeng:
if not define TURN_NO_TLS in file Configure,  then coturn will supoort tls. In this case, if function "register_callback_on_ioa_socket" return -1, function "client_to_be_allocated_timeout_handler" will never be register, so the session will nerver be delete in the session map.

huhaipeng:
Actually, I think if "register_callback_on_ioa_socket" return -1,  the session will never be deleted, so the memory will leak

huhaipeng:
@ggarber

huhaipeng:
![image](https://user-images.githubusercontent.com/23566147/185833125-82dfd90a-80ad-4e8d-98b1-b5cf619237d4.png)

huhaipeng:
![image](https://user-images.githubusercontent.com/23566147/185833307-52581b73-380e-4ebe-bfed-0c3fb3d77d45.png)

huhaipeng:
@ggarber @eakraly

eakraly:
Hi @huhaipeng ! Thank you for the report!
Do you think this is something you can submit a PR for?

huhaipeng:
> Hi @huhaipeng ! Thank you for the report! Do you think this is something you can submit a PR for?

Yes, I have submit a PR for fixing this bug. the commit as follow:
[fix memory leak of turn session]( https://github.com/coturn/coturn/pull/956)
@eakraly

huhaipeng:
I have submit an another  PR for fixing this bug. As follow:
fix turn session leak #962
@eakraly @ggarber

ggarber:
Thank you very much! Merged now.

PR #956, opened 2022-08-24
fix memory leak of turn_session

huhaipeng:
fix bug mentioned in issue #950

### DIFF
--- a/src/apps/relay/netengine.c
+++ b/src/apps/relay/netengine.c
@@ -829,7 +829,8 @@ static int handle_relay_message(relay_server_handle rs, struct message_to_relay 
 			} else {
 				s->e = rs->ioa_eng;
 				if(open_client_connection_session(&(rs->server), &(sm->m.sm))<0) {
-					IOA_CLOSE_SOCKET(s);
+					ts_ur_super_session *ss = sm->m.sm.s->session;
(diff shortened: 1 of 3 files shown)


eakraly:
Thanks @huhaipeng !

I have not tested it myself - could you please provide more detailed description of what objects are leaking, and how your fix addresses that?
Also, if you did some tests or even long running tests - what results did you see?

Thank you! (just trying to make sure we document enough information about the bugs and fixes so that it is easier to understand why those changes were done and how they fix the issues without the need to dive deep into the code)

huhaipeng:
> 

Actually, as below, before fixing, the session_map may be larger and larger when register_callback_on_ioa_socket inside open_client_connection_session retrun -1. The reason is that, if register_callback_on_ioa_socket return -1, no read/write/ event will be registered and no timeout event will be registered, in this case, the session will never be deleted from session_map.
image

After fixing, if open_client_connection_session return -1, shutdown_client_connection will be called, the session created from open_client_connection_session will be deleted and free in shutdown_client_connection.
@eakraly @ggarber

huhaipeng:
I haven't writed any unit tests. But I have applied my fix version in my test envrinment, it seems work correctly
@eakraly @ggarber

huhaipeng:
I found a better way to resolve this bug, close this PR.

PR #960, opened 2022-08-25
Use a single SSL context instead of 5 for TLS connections

eakraly:
All versions of TLS can be supported using a single SSL_CTX - setting minimal/maximal supported version should be done using, for example, `SSL_CTX_set_options(&turn_params.tls_ctx, SSL_OP_NO_TLSv1);` (or using `SSL_CTX_set_max_proto_version` for openssl-1.1.0 and newer)

`SSL_CTX_set_max_proto_version` is not used (yet) and might introduce breakage if using openssl-1.1.1 and newer without API compatibility mode.

This is a first in series of changes to rework openssl usage.

Links:
https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_options.html
https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_max_proto_version.html

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -83,13 +83,7 @@ char HTTP_ALPN[128] = "http/1.1";
 #define DEFAULT_GENERAL_RELAY_SERVERS_NUMBER (1)
 
 turn_params_t turn_params = {
-NULL, NULL,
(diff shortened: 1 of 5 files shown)


eakraly:
I'll reopen it after some cleanup

PR #955, opened 2022-08-24
Improve openssl3 and FIPS support

eakraly:
openssl-3.0 deprecated some APIs and introduced new APIs instead:

`SSL_get_peer_certificate ` -> `SSL_get1_peer_certificate `
`FIPS_mode()`->`EVP_default_properties_is_fips_enabled()`
`EVP_MD_CTX_set_flags()`->`EVP_default_properties_enable_fips()` specifically for enabling FIPS mode

This change should workaround that by ifdef-ing old/new versions of openssl and APIs - so pre-3.0 use existing APIs (so not change there) and >=3.0 will use new APIs (whether it actually works or not is still TBD as this is just a first step in openssl-3.0 support)

Should fix #886

Test Plan:
Run CI build that supports ubuntu-20.04 (openssl-1.1.1) and ubuntu-22.04 (openssl-3.0.2)
Both builds pass
None of them have FIPS support (which for 1.1.x stays the same as before)

PR #961, opened 2022-08-25
Add build and test for openssl-3.0

eakraly:
Add GitHub action that builds and runs tests using openssl-3.0 (default version in ubuntu 22.04)

It is expected to fail until openssl-3.0 compatibility is introduced (#955)

A change in cmake action is to make sure we actually run ubuntu:20.04 (which uses openssl-1.1.1). Otherwise, GH will change the version of ubuntu-latest at some point and it might not be very visible


ggarber:
Thx! can we keep a single job with a "matrix" to avoid the duplication?

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

eakraly:
Sure, I was looking for support of something similar but missed it obviously

ggarber (approved):
Thank you!

PR #962, opened 2022-08-26
fix turn session leak

huhaipeng:
Actually, as below, before fixing, the sessions_map may be larger and larger when register_callback_on_ioa_socket inside open_client_connection_session retrun -1. The reason is that, if register_callback_on_ioa_socket return -1, no read/write/ event will be registered and no timeout event will be registered, in this case, the session will never be deleted from sessions_map.
image

After fixing, when register_callback_on_ioa_socket fail，it doesn't return immediately, instead, we still register client_to_be_allocated_timeout_handler in the next step of open_client_connection_session. In this case, the unuseful session will be deleted and freed after timeout. 

@eakraly @ggarber

### DIFF
--- a/src/server/ns_turn_server.c
+++ b/src/server/ns_turn_server.c
@@ -4719,7 +4719,7 @@ static int attach_socket_to_session(turn_turnserver* server, ioa_socket_handle s
 
 int open_client_connection_session(turn_turnserver* server,
 				struct socket_message *sm) {
-
+	int ret = 0;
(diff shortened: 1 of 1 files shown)


ggarber:
Thank you very much @huhaipeng .

I looked at the code and your fix looks reasonable and safe.    I will merge it unless @eakraly has anything against it.

eakraly:
This looks like a much better fix for the issue @huhaipeng 

PR #967, opened 2022-09-02
Fix small issues reported by cppcheck

eakraly:
- Redundant checks for variable values
- Potential nullptr dereference
- Double check for the same variable

Run with:
```
cppcheck --inline-suppr --language=c --enable=warning,performance --force .
```

### DIFF
--- a/src/apps/common/hiredis_libevent2.c
+++ b/src/apps/common/hiredis_libevent2.c
@@ -240,7 +240,11 @@ redis_context_handle redisLibeventAttach(struct event_base *base, char *ip0, int
 	  port=port0;
 
   ac = redisAsyncConnect(ip, port);
-  if (ac->err) {
+  if (!ac) {
+  	fprintf(stderr,"Error: redisAsyncConnect returned NULL\n");
+  	return NULL;
+  }
+  else if (ac->err) {
   	fprintf(stderr,"Error: %s:%s\n", ac->errstr, ac->c.errstr);
   	return NULL;
   }
--- a/src/apps/common/hiredis_libevent2.c
+++ b/src/apps/common/hiredis_libevent2.c
@@ -240,7 +240,7 @@ redis_context_handle redisLibeventAttach(struct event_base *base, char *ip0, int
 	  port=port0;
 
   ac = redisAsyncConnect(ip, port);
-  if (!ac) {
+  if (ac->err) {
   	fprintf(stderr,"Error: %s:%s\n", ac->errstr, ac->c.errstr);
   	return NULL;
   }
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -2451,7 +2451,7 @@ int main(int argc, char **argv)
 		}
         }
 
-	if(use_cli && cli_password[0]==0 && use_cli) {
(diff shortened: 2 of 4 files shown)


src/apps/relay/mainrelay.c
```diff
@@ -2451,7 +2451,7 @@ int main(int argc, char **argv)
 		}
         }
 
-	if(use_cli && cli_password[0]==0 && use_cli) {
```
eakraly, line 2454: Double check `use_cli `

src/apps/relay/mainrelay.c
```diff
@@ -2753,8 +2753,7 @@ static void adjust_key_file_name(char *fn, const char* file_title, int critical)
 	  strncpy(fn,full_path_to_file,sizeof(turn_params.cert_file)-1);
 	  fn[sizeof(turn_params.cert_file)-1]=0;
 
-	  if(full_path_to_file)
```
eakraly, line 2756: full_path_to_file cannot be null here - checked before

src/apps/relay/tls_listener.c
```diff
@@ -66,16 +66,18 @@ static void server_input_handler(struct evconnlistener *l, evutil_socket_t fd,
 
 	tls_listener_relay_server_type * server = (tls_listener_relay_server_type*) arg;
 
+	if (!server)
```
eakraly, line 69: Must be checked here before next if

src/apps/relay/tls_listener.c
```diff
@@ -133,16 +135,18 @@ static void sctp_server_input_handler(struct evconnlistener *l, evutil_socket_t
...
```
eakraly, line 138: Should be checked here before dereferencing it next

src/client/ns_turn_msg.c
```diff
@@ -837,7 +837,7 @@ int stun_is_channel_message_str(const uint8_t *buf, size_t *blen, uint16_t* chnu
 
 			if(mandatory_padding) {
 				return 0;
-			} else if ((datalen_actual < datalen_header) || (datalen_header == 0)) {
```
eakraly, line 840: `(datalen_actual < datalen_header` already tested above

src/apps/common/hiredis_libevent2.c
```diff
@@ -240,7 +240,7 @@ redis_context_handle redisLibeventAttach(struct event_base *base, char *ip0, int
 	  port=port0;
 
   ac = redisAsyncConnect(ip, port);
-  if (!ac) {
+  if (ac->err) {
```
ggarber, line 243: What if ac is NULL?  Checking the `redisAsyncConnect` it looks like it can return NULL in a couple of cases

https://github.com/redis/hiredis/blob/master/async.c#L172

ggarber (changes requested):

src/apps/common/hiredis_libevent2.c
eakraly: You are right. Missed that

eakraly: I addressed this comment - they are not handled one after another

ggarber (approved):

PR #981, opened 2022-09-14
Add CI tests in older ubuntu version

ggarber:
This should help catching issues like the one in #978 with older versions of openssl or other libraries used

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,19 +11,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


ggarber:
I think it is not possible to run in ubuntu-16 anymore with GA and I will need to use docker for this.   Will try to do it.

eakraly:
> I think it is not possible to run in ubuntu-16 anymore with GA and I will need to use docker for this. Will try to do it.
Sorry I did not mention it - it would have been much easier if they still had it.

eakraly:
Nice! Container inside container

PR #983, opened 2022-09-14
Preserve file timestamps when using install(1)

robert-scheck:
- Preserve file timestamps when using `install`
- Use permissions `0644` rather than default `0755` for installing man pages

The alternative calls of `cp` are using `cp -p` as well (if `install` is unavailable).

### DIFF
--- a/configure
+++ b/configure
@@ -461,12 +461,12 @@ else
 	type install 2>>/dev/null
 	ER=$?
 	if [ ${ER} -eq 0 ] ; then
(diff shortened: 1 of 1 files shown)


robert-scheck:
CI failure `coturn/src/client/ns_turn_msg.c:260: undefined reference to 'FIPS_mode'` is _not_ related to my pull request.

eakraly (approved):

PR #989, opened 2022-09-16
Use a single SSL context object

eakraly:
openssl allows multiple TLS version support through a single SSL_CTX object.

This PR replaces 4 per-version SSL_CTX objects with a single object (DTLS is not yet changed).
SSL context initialization code for openssl with modern API (>=1.1.0) uses  `TLS_server_method` and `SSL_CTX_set_min_proto_version` instead of enabling specific TLS version. Byproduct of this is TLSv1_3 support when used with openssl-1.1.1 and above

TLS 1.2 and TLS 1.3 cannot be disabled (as before)

Test plan:
- run_tests.sh script now runs turnserver with SSL certificate (which enables TLS support)
- run_tests.sh now has one more basic test that uses TLS protocol


examples/run_tests.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 echo 'Running turnserver'
-../bin/turnserver --use-auth-secret  --static-auth-secret=secret --realm=north.gov --allow-loopback-peers --no-cli > /dev/null &
+../bin/turnserver --use-auth-secret  --static-auth-secret=secret --realm=north.gov --allow-loopback-peers --no-cli --cert ../examples/ca/turn_server_cert.pem --pkey ../examples/ca/turn_server_pkey.pem > /dev/null &
```
eakraly, line 4: Run turnserver with SSL certificate to support TLS

examples/run_tests.sh
```diff
@@ -16,6 +16,15 @@ else
 	exit $?
 fi
 
+echo 'Running turn client TLS'
```
eakraly, line 19: Run uclient with TLS to test basic TLS support

src/apps/relay/mainrelay.c
```diff
@@ -83,13 +83,7 @@ char HTTP_ALPN[128] = "http/1.1";
 #define DEFAULT_GENERAL_RELAY_SERVERS_NUMBER (1)
 
 turn_params_t turn_params = {
-NULL, NULL,
```
eakraly, line 86: Instead 4 SSL_CTX objects we now have 1

src/apps/relay/mainrelay.c
```diff
@@ -3148,27 +3142,6 @@ static void set_ctx(SSL_CTX** out, const char *protocol, const SSL_METHOD* metho
 	{
 		int op = 0;
 
-#if defined(SSL_OP_NO_SSLv2)
```
eakraly, line 3151: Support for TLS versions here was duplicated
Only set in one place (so removing here)

ggarber (commented):
Good job!

If understand it correctly this PR makes 2 backwards incompatible changes:

1/ SSL_OP_NO_SSLv2 is not disabled anymore.   It looks like in some old openssl versions it was supported and needed.

2/ Before it was possible to disable TLS 1.2 but keep 1.1 and 1.3 enabled and with this change that's not supported anymore.  I'm not saying that was useful, just trying to confirm if the behaviour changed and how the doc needs to be updated correspondingly if we really want to do that.

eakraly:
Thank you!

1/ This is correct. Let me add this - I missed the point in older version of openssl.
2/ Correct. I think this is the right way to go (treating it as minimal TLS version and there is no way around it). +1 to adding documentation.

Issue #118, opened 2017-04-05
CentOS 7 - CoTURN v4.5.0.6 wont run


shamun:
CentOS 7, Linux localhost.localdomain 3.10.0-514.6.2.el7.x86_64, CoTURN v4.5.0.6

**Step 1**  - Download & Install http://turnserver.open-sys.org/downloads/v4.5.0.6/
**Step 2** - Run/Start

`systemctl restart turnserver.service`

**Step 3** -  FAIL

```
[root@localhost ~]# systemctl restart turnserver.service
Job for turnserver.service failed because a configured resource limit was exceeded. See "systemctl status turnserver.service" and "journalctl -xe" for details.
[root@localhost ~]# systemctl status turnserver.service
● turnserver.service - coturn
   Loaded: loaded (/usr/lib/systemd/system/turnserver.service; disabled; vendor preset: disabled)
   Active: failed (Result: resources) since Wed 2017-04-05 08:50:26 CEST; 15s ago
     Docs: man:coturn(1)
           man:turnadmin(1)
           man:turnserver(1)
  Process: 2460 ExecStart=/usr/bin/turnserver -o -c /etc/turnserver/turnserver.conf $EXTRA_OPTIONS (code=exited, status=0/SUCCESS)

Apr 05 08:50:25 localhost.localdomain turnserver[2465]: 0: Trying to bind fd 18 to <127.0.0.1:80>: errno=13
Apr 05 08:50:25 localhost.localdomain turnserver[2465]: 0: Cannot bind TLS/TCP listener socket to addr 127.0.0.1:80
Apr 05 08:50:25 localhost.localdomain turnserver[2465]: 0: Trying to bind TLS/TCP listener socket to addr 127.0.0.1:80, again...
Apr 05 08:50:26 localhost.localdomain turnserver[2465]: 0: Trying to bind fd 16 to <127.0.0.1:80>: errno=13
Apr 05 08:50:26 localhost.localdomain turnserver[2465]: 0: Cannot bind DTLS/UDP listener socket to addr 127.0.0.1:80
Apr 05 08:50:26 localhost.localdomain turnserver[2465]: 0: Fatal final failure: cannot bind DTLS/UDP listener socket to addr 127.0.0.1:80
Apr 05 08:50:26 localhost.localdomain systemd[1]: turnserver.service never wrote its PID file. Failing.
Apr 05 08:50:26 localhost.localdomain systemd[1]: Failed to start coturn.
[...]

PR #990, opened 2022-09-19
Fix issue #118 by changing systemd service type from notify to exec

rapsys:
The first commit fix the issue #118 that prevent the process to start as coturn never send the notification to systemd thus preventing the startup.

Secont commit fix the default pid path to conform to Filesystem Hierarchy Standard

### DIFF
--- a/README.turnserver
+++ b/README.turnserver
@@ -564,7 +564,7 @@ Options with values:
 			fits the address, then it is denied.
 
 --pidfile 		File name to store the pid of the process.
-			Default is /var/run/turnserver.pid (if superuser account is used) or
+			Default is /run/turnserver/turnserver.pid (if superuser account is used) or
 			/var/tmp/turnserver.pid .
 
 --acme-redirect  <URL>	Redirect ACME/RFC8555 (like Let's Encrypt challenge) requests, i.e.
--- a/docker/coturn/turnserver.conf
+++ b/docker/coturn/turnserver.conf
@@ -622,10 +622,10 @@ syslog
 # allowed-peer-ip=83.166.68.45
 
 # File name to store the pid of the process.
-# Default is /var/run/turnserver.pid (if superuser account is used) or
+# Default is /run/turnserver/turnserver.pid (if superuser account is used) or
 # /var/tmp/turnserver.pid .
 #
-#pidfile="/var/run/turnserver.pid"
+#pidfile="/run/turnserver/turnserver.pid"
 
 # Require authentication of the STUN Binding request.
 # By default, the clients are allowed anonymous access to the STUN Binding functionality.
--- a/examples/etc/turnserver.conf
+++ b/examples/etc/turnserver.conf
@@ -649,10 +649,10 @@
 # allowed-peer-ip=83.166.68.45
 
 # File name to store the pid of the process.
-# Default is /var/run/turnserver.pid (if superuser account is used) or
+# Default is /run/turnserver/turnserver.pid (if superuser account is used) or
 # /var/tmp/turnserver.pid .
 #
-#pidfile="/var/run/turnserver.pid"
+#pidfile="/run/turnserver/turnserver.pid"
 
 # Require authentication of the STUN Binding request.
 # By default, the clients are allowed anonymous access to the STUN Binding functionality.
--- a/man/man1/turnserver.1
+++ b/man/man1/turnserver.1
@@ -812,7 +812,7 @@ fits the address, then it is denied.
 .B
 \fB\-\-pidfile\fP
 File name to store the pid of the process.
-Default is /var/run/turnserver.pid (if superuser account is used) or
+Default is /run/turnserver/turnserver.pid (if superuser account is used) or
 /var/tmp/turnserver.pid .
 .TP
 .B
--- a/rpm/turnserver.init.el
+++ b/rpm/turnserver.init.el
@@ -6,7 +6,7 @@
 # description: RFC 5766 TURN Server
 #
 # processname: turnserver
-# pidfile: /var/run/turnserver/turnserver.pid
+# pidfile: /run/turnserver/turnserver.pid
 # config: /etc/turnserver/turnserver.conf
 #
 ### BEGIN INIT INFO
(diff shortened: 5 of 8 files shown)


ggarber:
Thank you very much @rapsys .   Are you sure this will work in all the supported platforms?

For example I don't have a /run in MacOS but I have the /var/run.

rapsys:
> Thank you very much @rapsys . Are you sure this will work in all the supported platforms?
> 
> For example I don't have a /run in MacOS but I have the /var/run.

I have no way to test it on MacOS too poor for such device with my old 10 years old laptop :p

I will look to revert it for mac then in an other patch when I have time.

rapsys:
I looked at my change again, it should work on MacOS by just adding the "pidfile /var/run/turnserver/turnserver.pid" directive to configuration.

There is no MacOS specific configuration or service configuration from what I saw.

ggarber:
> I looked at my change again, it should work on MacOS by just adding the "pidfile /var/run/turnserver/turnserver.pid" directive to configuration.
> 
> There is no MacOS specific configuration or service configuration from what I saw.

If I compile and run the turnserver locally with this branch in my MacOS it is not able to write the pid because /run folder doesn't exist while before this change it is perfectly fine writting to /var/run.

Either we use different files for different OSs or we give up on conforming to the Filesystem Hierarchy Standard.   Any other alternative?

BTW, if you want to create a separate PR just for the "systemd service type" I can merge that part quickly.

rapsys:
I added a detection of linux >= 3.0.0 where /var/run moved to /run on the source code.

The switch was done in linux >= 3.0 if I consider this bug report :
https://bugs.linuxfoundation.org/show_bug.cgi?id=718

This way the macos, bsd and linux < 3.0.0 should be able to run the server with default config and it should be fine for everybody.

ggarber:
Thank you very much for looking at this @rapsys  but if you use KERNEL_VERSION that will be evaluated during compilation and not during execution.  So if you compile coturn in kernel 2.9 and then run it in 3.0 it will still use /var/run.


What is the problem of leaving the existing path?   I have checked randomly two very popular services (apache httpd and mysql server) and it looks like both still use /var/run.    Isn't that safer?

https://github.com/apache/httpd/blob/21f16155c38e406e0a0daaa60a539d66128cf044/build/rpm/httpd.init#L61

https://github.com/mysql/mysql-server/blob/3e90d07c3578e4da39dc1bce73559bbdf655c28c/packaging/rpm-common/mysqlrouter.init#L47


As I said we can land your first commit (systemd service type) quickly if you want to slit it.

rapsys:
> Thank you very much for looking at this @rapsys but if you use KERNEL_VERSION that will be evaluated during compilation and not during execution. So if you compile coturn in kernel 2.9 and then run it in 3.0 it will still use /var/run.

Linux kernel version 3.0 went out in July 2011 and it's end of life happened in October 2013 :
https://en.wikipedia.org/wiki/Linux_kernel_version_history#Releases_3.x.y

Ubuntu 11.04 is end of life since October 2012 with kernel < 3.0, same for all major distributions.

The only distribution to consider is RHEL 6, currently in extended life-cycle support until june 2024, but when you build for a RHEL, you will build on the same version to have the same library dependancies :
https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux#RHEL_6

Usualy linux build systems will install in a chroot all the required package from the target distribution, thus the kernel version will be right, the one from the target, not the one from uname -a.

There is not need to have this detection on startup.

robert-scheck:
I'm the RPM package maintainer for coturn in Fedora and EPEL (for CentOS, RHEL and Rocky Linux), and there we use our own systemd unit file. Building coturn for RHEL 6 (ELS phase until 2024) can IMHO be considered out of target, because of a) SysV initscripts rather than systemd and b) it's already challenging to build coturn for EPEL 7 due to old software versions around. Thus no need for run-time checks from my point of view.

eakraly:
Adding my 2 cents:

1. We all agree that 1st commit is good to go (Change from type notify to exec). We should split it out (atomic commits FTW)

3. PID file location: is there a real need to change default (across so many files) if turnserver would run through the list until succeeds? We can add the entries and change the order to make sure turnserver.pid is created in the "right" place first.

rapsys:
> On top of that, going forward, running in a container is going to be more common way to run turnserver (my personal opinion and I am biased - my system runs turnserver in containers in k8s). With containers there is no need to have pid file (even worse - having pid file prevents from using read-only file system which is great security tool) see #994
> 
> (having readonly root file system is a recommended best practice. for example https://aws.github.io/aws-eks-best-practices/security/docs/pods/#configure-your-images-with-read-only-root-file-system)

I have no idea how those containers works, there is not even a /tmp or /run writeable and not persistent ?

I would suggest to refuse to start unless a pid is dropped either at location specified in config or default locations and limit to only (/var)?/run/turnserver/turnserver.pid then (/var)?/run/turnserver.pid.

For container you should set it to /dev/null for example.

rapsys:
> Actually, why not pass extra `--pidfile /run/turnserver/turnserver.pid` in relevant file `rpm/turnserver.service.fc` ?
> 
> Would that not achieve the desired result (without changing code/documentation)?

That is not a clean solution, a daemon must drop a pid file and if not able to must not start as there is no reliable way to act on it else.

Start a daemon in readonly filesystem is a borderline case, usualy there is always various way to escape a chroot/jail/etc and access writable paths (/tmp, /var/tmp, /run/turnserver, etc...) through various exploit...

People doing that must specify /dev/null as pidfile either as configuration or runtime parameter.

With systemd you may configure a service which don't need pidfile as it will get the pid on the fork call, but it don't cover every cases.

In my opinon we are drifting away from this bug report issue concerning the systemd invalid service file and default pidfile paths.

rapsys:
And adding a runtime parameter to a service file is kind of crappy. Who wants to have useless runtime parameters on a process which may have been configured correctly as default at build time ?

ggarber:
Thank you @rapsys but the way the PR is is not correct.    In the README and the configuration files it says "Default is /run/turnserver/turnserver.pid" but that's only the case in some OSs.

Can you please leave in this PR only the changes related to the title (Fix issue #118) so that we can merge it quickly given that apparently there is no consensus or simple solution for the second part?

Thank you very much and apologize for the inconvenience

PR #995, opened 2022-09-25
Silence warnings by converting STRCPY to strncpy calls

eakraly:
STRCPY macro makes pointer comparison which creates a warning

In those places, replace the macro with `strncpy` with careful review of destination buffer size


src/apps/common/hiredis_libevent2.c
```diff
@@ -233,7 +233,7 @@ redis_context_handle redisLibeventAttach(struct event_base *base, char *ip0, int
   if(ip0 && ip0[0])
 	  STRCPY(ip,ip0);
```
ggarber, line 234: isn't it here where strncpy matters and not 2 lines below where the IP is fixed lenght?

Same in the next file

src/apps/relay/turn_admin_server.c
```diff
@@ -1557,7 +1557,8 @@ static ioa_socket_handle current_socket = NULL;
 static char *get_bold_admin_title(void)
 {
 	static char sbat[1025];
-	STRCPY(sbat,__bold_admin_title);
+	strncpy(sbat,__bold_admin_title,TURN_LONG_STRING_SIZE);
```
ggarber, line 1560: sizeof(sbat) looks safer

or alternatively use TURN_LONG_STRING_SIZE also in "sbat" definition, but i think the sizeof is safer.


(This applies also to the previous file mainrelay.c)

src/apps/common/hiredis_libevent2.c
```diff
@@ -233,7 +233,7 @@ redis_context_handle redisLibeventAttach(struct event_base *base, char *ip0, int
...
   else
-	  STRCPY(ip,"127.0.0.1");
+	  strncpy(ip,"127.0.0.1",256);
```
ggarber, line 236: use sizeof(ip)

ggarber:
STRCPY macro was making sure that the final string was \0 terminated while the new code breaks that assumption.  Is not an issue?

Shouldn't we extend the STRCPY to STRNCPY macro instead?

eakraly:
> STRCPY macro was making sure that the final string was \0 terminated while the new code breaks that assumption. Is not an issue?
strncpy also copies the end `\0` from the source but good 

> Shouldn't we extend the STRCPY to STRNCPY macro instead?
STRCPY macro is actually calling strncpy internally. And doing a few other optimizations (some questionable I would say).

ggarber:
> > STRCPY macro was making sure that the final string was \0 terminated while the new code breaks that assumption. Is not an issue?
> 
> strncpy also copies the end `\0` from the source but good point - I will check those specific places again for validity

@eakraly Can you double check?  I don't think that is the case when the length is the maximum one (source length = destination length = size)

src/apps/common/hiredis_libevent2.c
eakraly: I was just going after warnings this time
The warning is generated in STRCPY(dst,"string") and not in STRCPY(dst,src)

eakraly:
> @eakraly Can you double check? I don't think that is the case when the length is the maximum one (source length = destination length = size)

What STRCPY macro does (and that is why I say it is questionable) is that it may copy the full size source string and then overwrite last character with `\0`

with "just" strncpy it will copy last '\0' if the source string contains it. I will check if those specific cases where I changed there is any risk of that happenning.

src/apps/common/hiredis_libevent2.c
```diff
@@ -233,7 +233,7 @@ redis_context_handle redisLibeventAttach(struct event_base *base, char *ip0, int
   if(ip0 && ip0[0])
 	  STRCPY(ip,ip0);
   else
-	  STRCPY(ip,"127.0.0.1");
+	  strncpy(ip,"127.0.0.1",sizeof(ip));
```
eakraly, line 236: "127.0.0.1" is null-terminated so strncpy will copy the null
Longest IP representation is 45 https://stackoverflow.com/a/7477384

src/apps/relay/dbdrivers/dbd_redis.c
```diff
@@ -189,7 +189,7 @@ redis_context_handle get_redis_async_connection(struct event_base *base, const c
 				if (co->host)
 					STRCPY(ip,co->host);
 				if (!ip[0])
-					STRCPY(ip,"127.0.0.1");
+					strncpy(ip,"127.0.0.1",sizeof(ip));
```
eakraly, line 192: same as above

src/apps/relay/dbdrivers/dbd_redis.c
```diff
@@ -298,7 +298,7 @@ static redisContext *get_redis_connection(void) {
 			if (co->host)
 				STRCPY(ip,co->host);
 			if (!ip[0])
-				STRCPY(ip,"127.0.0.1");
+				strncpy(ip,"127.0.0.1",sizeof(ip));
```
eakraly, line 302: same as above

ggarber:
> with "just" strncpy it will copy last '\0' if the source string contains it.

If the source contains it AND it fits in the max length.

> I will check if those specific cases where I changed there is any risk of that happenning.

Yep, that's the problem I see. Without the macro you depend on specific lengths of specific strings and you need to check them one by one to know if the code is correct or will crash.

Anyway approved in case you think that's ok and you want to land it.

ggarber (approved):

src/apps/relay/mainrelay.c
```diff
@@ -2075,7 +2075,7 @@ static int adminmain(int argc, char **argv)
 
 #if !defined(TURN_NO_SQLITE)
 	if(!strlen(turn_params.default_users_db.persistent_users_db.userdb) && (turn_params.default_users_db.userdb_type == TURN_USERDB_TYPE_SQLITE))
-		STRCPY(turn_params.default_users_db.persistent_users_db.userdb,DEFAULT_USERDB_FILE);
+        strncpy(turn_params.default_users_db.persistent_users_db.userdb,DEFAULT_USERDB_FILE, TURN_LONG_STRING_SIZE);
```
eakraly, line 2078: Const null-terminated string "TURNDB"

src/apps/relay/mainrelay.c
```diff
@@ -2426,7 +2426,7 @@ int main(int argc, char **argv)
 
 #if !defined(TURN_NO_SQLITE)
 	if(!strlen(turn_params.default_users_db.persistent_users_db.userdb) && (turn_params.default_users_db.userdb_type == TURN_USERDB_TYPE_SQLITE))
-			STRCPY(turn_params.default_users_db.persistent_users_db.userdb,DEFAULT_USERDB_FILE);
+            strncpy(turn_params.default_users_db.persistent_users_db.userdb,DEFAULT_USERDB_FILE, TURN_LONG_STRING_SIZE);
```
eakraly, line 2429: Const null-terminated string "TURNDB"

src/apps/relay/mainrelay.c
```diff
@@ -2979,7 +2979,7 @@ static void set_ctx(SSL_CTX** out, const char *protocol, const SSL_METHOD* metho
 	SSL_CTX_set_default_passwd_cb(ctx, pem_password_func);
 
 	if(!(turn_params.cipher_list[0]))
-		STRCPY(turn_params.cipher_list,DEFAULT_CIPHER_LIST);
+        strncpy(turn_params.cipher_list,DEFAULT_CIPHER_LIST,TURN_LONG_STRING_SIZE);
```
eakraly, line 2982: Const null-terminated string "DEFAULT"

src/apps/relay/turn_admin_server.c
```diff
@@ -1557,7 +1557,8 @@ static ioa_socket_handle current_socket = NULL;
 static char *get_bold_admin_title(void)
 {
 	static char sbat[1025];
-	STRCPY(sbat,__bold_admin_title);
+	strncpy(sbat,__bold_admin_title,sizeof(sbat));
```
eakraly, line 1560: Const null terminated string "<b>TURN Server</b><br><i>https admin connection</i><br>\r\n"

PR #996, opened 2022-09-26
Use single SSL_CTX for DTLS support

### LINKED ISSUE #924: Build 4.5.2-r11 from source with openssl1.1.1q failure because some methods referred by the source code are deprecated  in openssl 1.1.1
We try to build 4.5.2-r11 from source code, but failed because some of the openssl method used in 4.5.2-r11 are deprecated in openssl  version 1.1.1 now, for example, in the fiile of src\apps\relay\mainrelay.c, line#3219-3255, it uses following method 
TLSv1_1_server_method()
TLSv1_2_server_method()
DTLSv1_2_server_method()
DTLSv1_server_method()

While those methods are deprecated in openssl version 1.1.1 according to the document https://www.openssl.org/docs/man1.1.1/man3/TLSv1_1_method.html

And also in the openssl/ssl.h, they are defined as deprecated, so when compile 4.5.2-r11 with openssl 1.1.1q, the build is failed

# ifndef OPENSSL_NO_TLS1_1_METHOD
DEPRECATEDIN_1_1_0(__owur const SSL_METHOD *TLSv1_1_method(void)) /* TLSv1.1 */
DEPRECATEDIN_1_1_0(__owur const SSL_METHOD *TLSv1_1_server_method(void))
DEPRECATEDIN_1_1_0(__owur const SSL_METHOD *TLSv1_1_client_method(void))
# endif

# ifndef OPENSSL_NO_TLS1_2_METHOD
DEPRECATEDIN_1_1_0(__owur const SSL_METHOD *TLSv1_2_method(void)) /* TLSv1.2 */
DEPRECATEDIN_1_1_0(__owur const SSL_METHOD *TLSv1_2_server_method(void))
DEPRECATEDIN_1_1_0(__owur const SSL_METHOD *TLSv1_2_client_method(void))
# endif

# ifndef OPENSSL_NO_DTLS1_METHOD
DEPRECATEDIN_1_1_0(__owur const SSL_METHOD *DTLSv1_method(void)) /* DTLSv1.0 */
DEPRECATEDIN_1_1_0(__owur const SSL_METHOD *DTLSv1_server_method(void))
DEPRECATEDIN_1_1_0(__owur const SSL_METHOD *DTLSv1_client_method(void))
# endif

should the CoTURN code be changed to  replaced these method by TLS_method(), TLS_server_method(), TLS_client_method() DTLS_method(), DTLS_server_method(), DTLS_client_method() etc?

eakraly:
Similar to #989, use a single SSL context for all versions of DTLS protocol

- Add support for modern API (protocol version independent APIs)
- Add DTLS test to the CI test
- Removing calls to `SSL_CTX_set_read_ahead` in DTLS context (does nothing as DTLS is datagram protocol - we always get the whole datagram so this call has no impact)

Fixes #924

### DIFF
--- a/src/apps/relay/dtls_listener.c
+++ b/src/apps/relay/dtls_listener.c
@@ -279,18 +279,7 @@ static ioa_socket_handle dtls_server_input_handler(dtls_listener_relay_server_ty
 	timeout.tv_usec = 0;
 	BIO_ctrl(wbio, BIO_CTRL_DGRAM_SET_RECV_TIMEOUT, 0, &timeout);
 
-#if DTLSv1_2_SUPPORTED
-	if(get_dtls_version(ioa_network_buffer_data(nbh),
-							(int)ioa_network_buffer_get_size(nbh)) == 1) {
-		connecting_ssl = SSL_new(server->e->dtls_ctx_v1_2);
-	} else {
-		connecting_ssl = SSL_new(server->e->dtls_ctx);
-	}
-#else
-	{
-		connecting_ssl = SSL_new(server->e->dtls_ctx);
-	}
-#endif
+	connecting_ssl = SSL_new(server->e->dtls_ctx);
 
 	SSL_set_accept_state(connecting_ssl);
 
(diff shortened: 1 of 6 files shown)


src/apps/relay/dtls_listener.c
```diff
@@ -966,8 +944,6 @@ void setup_dtls_callbacks(SSL_CTX *ctx) {
   SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER | SSL_VERIFY_CLIENT_ONCE, dtls_verify_callback);
 #endif
 
-  SSL_CTX_set_read_ahead(ctx, 1);
```
ggarber, line 969: afaik this is fine but can you add to the PR description why is this removed / not needed for future reference?

src/apps/relay/mainrelay.c
```diff
@@ -3260,24 +3246,32 @@ static void openssl_load_certificates(void)
...
+		set_ctx(&turn_params.dtls_ctx,"DTLS",DTLSv1_2_server_method()); // openssl-1.0.2
+        if(!turn_params.no_tlsv1_2) {
+            SSL_CTX_set_options(turn_params.dtls_ctx, SSL_OP_NO_DTLSv1_2);
+        }
 #else
-		set_ctx(&turn_params.dtls_ctx,"DTLS",DTLSv1_server_method());
+        set_ctx(&turn_params.dtls_ctx,"DTLS",DTLSv1_server_method()); // < openssl-1.0.2
 #endif
-		SSL_CTX_set_read_ahead(turn_params.dtls_ctx, 1);
+        if(!turn_params.no_tlsv1) {
```
ggarber, line 3261: what about no-tlsv1_1 ?  it looks like if we are checking only for no_tlsv1_2 and no_tlsv1

ggarber (commented):
thx!

src/apps/relay/mainrelay.c
eakraly: Right. I think it also how it worked before (there is no DTLS1.1)

ggarber:
Thank you!

PR #997, opened 2022-09-27
Use epoll for promhttp server if supported. 

fancycode:
Before:
```
./bin/turnserver --prometheus -m 32
...
Socket descriptor larger than FD_SETSIZE: 2660 > 1024
0: : ERROR: Could not start Prometheus collector!
```

After:
```
./bin/turnserver --prometheus -m 32
...
0: : Prometheus collector started successfully.
```

Also added `MHD_USE_ERROR_LOG` so the actual problems during startup of promhttp are printed. Without it the `Socket descriptor larger than FD_SETSIZE: 2660 > 1024` error was not visible and finding the cause of the startup problems of promhttp was not obvious at first.

### DIFF
--- a/src/apps/relay/prom_server.c
+++ b/src/apps/relay/prom_server.c
@@ -70,7 +70,7 @@ int start_prometheus_server(void){
 
   promhttp_set_active_collector_registry(NULL);
 
-  unsigned int flags = MHD_USE_DUAL_STACK;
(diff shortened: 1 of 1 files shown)


eakraly:
Thank you for the contribution @fancycode !

eakraly (approved):

PR #1006, opened 2022-10-03
Malformed response to mobility refresh request

purehyperbole:
Hello, and thank you for this great project!

When sending a mobility refresh request, the server is responding with an invalid stun packet. Specifically, only the last 24 bytes generated for message integrity are sent.

Request:
image

Response:
image

This appears to be caused by the buffer length not being updated when `STUN_ATTRIBUTE_SOFTWARE` is added to the response, as it's being assigned to a `len` variable scoped to that conditional. 

As a result, the next call to `stun_get_command_message_len_str` when handling message integrity returns `-1` and resizes the buffer, etc.

Passing in the original `len` to `stun_attr_add_str` for attribute software fixes the issue, but apologies if this is not the right way to fix this!

### DIFF
--- a/src/server/ns_turn_server.c
+++ b/src/server/ns_turn_server.c
@@ -1770,7 +1770,6 @@ static int handle_turn_refresh(turn_turnserver *server,
 										{
 											const uint8_t *field = (const uint8_t *) get_version(server);
 											size_t fsz = strlen(get_version(server));
-											size_t len = ioa_network_buffer_get_size(nbh);
 											stun_attr_add_str(ioa_network_buffer_data(nbh), &len, STUN_ATTRIBUTE_SOFTWARE, field, fsz);
 											ioa_network_buffer_set_size(nbh, len);
 										}
--- a/src/server/ns_turn_server.c
+++ b/src/server/ns_turn_server.c
@@ -1770,16 +1770,19 @@ static int handle_turn_refresh(turn_turnserver *server,
 										{
 											const uint8_t *field = (const uint8_t *) get_version(server);
 											size_t fsz = strlen(get_version(server));
+											size_t len = ioa_network_buffer_get_size(nbh);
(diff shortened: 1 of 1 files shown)


ggarber:
Thank you for the patch.  It makes sense but i need to take a closer look at those methods.

Disabling the software attribute would solve it and it is a good practice if you need a workaround in the mean time.

Thx!

eakraly:
Is it something we can easily add a test to (in some form)?
Trying to think about coverage so we can (gradually) get more and more code covered by tests.

And thank you for contribution @purehyperbole !

ggarber:
@purehyperbole I took a look and this is handled in a different way in rest of places in turn (f.e Ln 4863 in ns_turnserver.c).

Find below what I think we should do in the code with the bug to be consistent with rest of the places.  Look at the "  // ADD THIS LINE" lines.   That way it always uses the "len" variable in the local scope and not the global one.

If you can update this PR and do some basic testing that would be great.  Let me know otherwise.

Thank you!

```
if(message_integrity) {
	size_t len = ioa_network_buffer_get_size(nbh);  // ADD THIS LINE
	stun_attr_add_integrity_str(server->ct,ioa_network_buffer_data(nbh),&len,ss->hmackey,ss->pwd,SHATYPE_DEFAULT);
	ioa_network_buffer_set_size(nbh,len);
}
if ((server->fingerprint) || ss->enforce_fingerprints) {
	size_t len = ioa_network_buffer_get_size(nbh);   // ADD THIS LINE
	if (stun_attr_add_fingerprint_str(ioa_network_buffer_data(nbh), &len) < 0) {
		*err_code = 500;
		ioa_network_buffer_delete(server->e, nbh);
		return -1;
	}
	ioa_network_buffer_set_size(nbh, len);
}
```

purehyperbole:
@ggarber Awesome, thanks for the feedback!

I've implemented the changes you requested and tested again. Looks to work OK!

ggarber:
Thank you very much @purehyperbole

PR #1009, opened 2022-10-06
Update turnserver.conf

hyorin:
Made some text proofreading

### DIFF
--- a/docker/coturn/turnserver.conf
+++ b/docker/coturn/turnserver.conf
@@ -1,7 +1,7 @@
 # Coturn TURN SERVER configuration file
 #
 # Boolean values note: where a boolean value is supposed to be used,
-# you can use '0', 'off', 'no', 'false', or 'f' as 'false,
+# you can use '0', 'off', 'no', 'false', or 'f' as 'false',
 # and you can use '1', 'on', 'yes', 'true', or 't' as 'true'
 # If the value is missing, then it means 'true' by default.
 #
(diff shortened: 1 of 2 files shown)


examples/etc/turnserver.conf
```diff
@@ -663,7 +664,7 @@
 #
 #mobility
 
-# Allocate Address Family according (DEPRECATED and will be removed in favour of allocation-default-address-family)
+# Allocate Address Family according (DEPRECATED and will be removed in favor of allocation-default-address-family)
```
eakraly, line 667: Technically not an issue (English vs American)

eakraly:
Thank you @hyorin !
Could you please do the same for `docker/coturn/turnserver.conf` file?

docker/coturn/turnserver.conf
```diff
@@ -723,3 +766,31 @@ cli-password=CHANGE_ME
...
+# Originally, if there are more than one listener address from the same
+# address family, then by default the NAT behavior discovery feature enabled.
+# This option disables the original behavior, because the NAT behavior
+# discovery adds extra attributes to response, and this increase the
+# possibility of an amplification attack.
+#
+# Strongly encouraged to use this option to decrease gain factor in STUN
+# binding responses.
+#
+no-rfc5780
```
eakraly, line 781: @ggarber this changes the default setting for docker containers
I do not think this is bad (better defaults - more "secure") - but it changes the default
Wanted to have another 👀 on it

ggarber: I think it is fine.    I think the docker images are new enough that the people using rfc5780 are probably not using them.

I would land it and include it in the "Release noteS".  I'll try to take anote.

PR #1012, opened 2022-10-08
Cleanup openssl initialization

eakraly:
Rewriting openssl initialization code (threading support to make it cleaner

- Regroup functions so that there is one ifdef (for old code and new code)
- Modern openssl (>1.0.2) does not need any synchornization routines so they are empty
- Old openssl (<=1.0.2) now require `OPENSSL_THREADS` which allows running multiple threads in turnserver. Not having turnserver multi-threaded is a huge waste. `OPENSSL_THREADS` is now a requirement.

Test Plan:
- CI builds pass for openssl versions 1.0.2, 1.1.1, 3.0, including tests

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -1345,7 +1345,6 @@ static void set_option(int c, char *value)
 		STRCPY(turn_params.relay_ifname, value);
 		break;
 	case 'm':
-#if defined(OPENSSL_THREADS)
 		if(atoi(value)>MAX_NUMBER_OF_GENERAL_RELAY_SERVERS) {
 			TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING, "WARNING: max number of relay threads is 128.\n");
 			turn_params.general_relay_servers_number = MAX_NUMBER_OF_GENERAL_RELAY_SERVERS;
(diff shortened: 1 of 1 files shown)


src/apps/relay/mainrelay.c
```diff
@@ -1345,7 +1345,6 @@ static void set_option(int c, char *value)
 		STRCPY(turn_params.relay_ifname, value);
 		break;
 	case 'm':
-#if defined(OPENSSL_THREADS)
```
eakraly, line 1348: Slowly chipping away compatibility with super-old openssl versions
Running turnserver in a single threaded mode using very old openssl versions is hard to support
Worst case we can branch away and support 4.6.x versions

src/apps/relay/mainrelay.c
```diff
@@ -2632,15 +2628,12 @@ int main(int argc, char **argv)
 
 ////////// OpenSSL locking ////////////////////////////////////////
 
-#if defined(OPENSSL_THREADS)
```
eakraly, line 2635: `OPENSSL_THREADS` only makes sense in 1.0.2 and below
aything above that do not need locking mechanisms

src/apps/relay/mainrelay.c
```diff
@@ -2632,15 +2628,12 @@ int main(int argc, char **argv)
...
-
-static char some_buffer[65536];
```
eakraly, line 2637: Random unused 64KB buffer, hmm

ggarber (commented):
I think with this change it will fail if openssl is not compiled with threading support.  I think that was the point of OPENSSL_THREADS macro.

In other words threading support not only depends on the version but also on the compilation options so it looks like the previous version was more robust in terms of supporting all the different options.

eakraly:
Correct mostly
Threading support does not apply to modern versions of openssl
But yes I agree - I was thinking about re-adding the check for OPENSSL_THREADS (even if for old versions)

Issue #272, opened 2018-09-11
admin interface user add logs mysql connection


misi:
turnadmin -a 

in function e.g. mysql_set_user_key
it is missing: donot_print_connection_success=1;

PR #1020, opened 2022-10-15
Sanitize DB connection string before printing to log

### LINKED ISSUE #1017: docker image coturn/coturn:debian is logging mysql db password on startup
Hi,
I think it is not a good idea to log the full password of db connection. Please remove following log or change it such that the password is not fully shown:
`0: : MySQL DB connection success: host=mariadb dbname=coturn user=coturn password=***** port=3306 connect_timeout=60 read_timeout=60`

eakraly:
Store sanitized version of DB connection string with password masked (replace all chars with * which exposes its length)
Use sanitized version when logging connection string

Fixes #1017 and #272

### DIFF
--- a/src/apps/relay/dbdrivers/dbdriver.c
+++ b/src/apps/relay/dbdrivers/dbdriver.c
@@ -111,7 +111,7 @@ const turn_dbdriver_t * get_dbdriver()
 	return _driver;
 }
 
-char* sanitize_users_db_string(char* udb) {
+char* sanitize_userdb_string(char* udb) {
     char * ret = NULL;
     char * pstart;
     char * pend;
--- a/src/apps/relay/dbdrivers/dbdriver.c
+++ b/src/apps/relay/dbdrivers/dbdriver.c
@@ -111,3 +111,39 @@ const turn_dbdriver_t * get_dbdriver()
 	return _driver;
 }
 
+char* sanitize_users_db_string(char* udb) {
+    char * ret = NULL;
+    char * pstart;
+    char * pend;
+
+    /* host=localhost dbname=coturn user=turn password=turn connect_timeout=30 */
+    ret = strdup(udb);
+    pstart = strstr (ret,"password=");
+    if (pstart != NULL)
+    {
+        pstart += strlen("password=");
+        pend = strstr (pstart," ");
+        size_t plen = pend - pstart;
+        if (pend == NULL)
+        {
+            plen = strlen(pstart);
+        }
+        memset(pstart,'*',plen);
+    }
+    else
+    {
+        /* postgresql://username:password@/databasename */
+        pstart = strstr (ret,"postgresql://");
+        if (pstart != NULL)
+        {
+            pstart += strlen("postgresql://");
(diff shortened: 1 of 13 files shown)


src/apps/relay/mainrelay.c
```diff
@@ -1501,40 +1501,65 @@ static void set_option(int c, char *value)
...
+        {
 #if defined(TURN_NO_SQLITE)
-	  TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING, "WARNING: Options -b, --userdb and --db are not supported because SQLite is not supported in this build.\n");
+          TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING, "WARNING: Options -b, --userdb and --db are not supported because SQLite is not supported in this build.\n");
 #else
-		STRCPY(turn_params.default_users_db.persistent_users_db.userdb, value);
-		turn_params.default_users_db.userdb_type = TURN_USERDB_TYPE_SQLITE;
+            STRCPY(turn_params.default_users_db.persistent_users_db.userdb, value);
+            turn_params.default_users_db.userdb_type = TURN_USERDB_TYPE_SQLITE;
+            char *s_userdb = sanitize_userdb_string(turn_params.default_users_db.persistent_users_db.userdb);
```
ggarber, line 1510: maybe this new code could go outside of the switch/case to avoid duplicating it 5 times.  Something like:

if (strlen(turn_params.default_users_db.persistent_users_db.userdb)) {
            char *s_userdb = sanitize_userdb_string(turn_params.default_users_db.persistent_users_db.userdb);
            STRCPY(turn_params.default_users_db.persistent_users_db.s_userdb, s_userdb);
            free(s_userdb);
}

src/apps/relay/dbdrivers/dbdriver.c
```diff
@@ -111,3 +111,39 @@ const turn_dbdriver_t * get_dbdriver()
...
+        if (pend == NULL)
+        {
+            plen = strlen(pstart);
+        }
+        memset(pstart,'*',plen);
+    }
+    else
+    {
+        /* postgresql://username:password@/databasename */
+        pstart = strstr (ret,"postgresql://");
```
ggarber, line 136: username is being replaced with '*' in this case.   I guess it is a known issue but just in case.    it lgtm anyway

ggarber (changes requested):
I think it is missing the same change in dbd_redis.c.

Rest of comments are not critical, just fyc if they make sense.

src/apps/relay/dbdrivers/dbdriver.c
eakraly: > I think it is missing the same change in dbd_redis.c.
> 
> Rest of comments are not critical, just fyc if they make sense.
My editor decided to skip the code text that is not enabled due to ifdef ...

eakraly:
> I think it is missing the same change in dbd_redis.c.
> 
> Rest of comments are not critical, just fyc if they make sense.
My editor decided to skip the code text that is not enabled due to ifdef ...

PR #1023, opened 2022-10-17
Reduce usage of TURN_NO_PROMETHEUS

eakraly:
`TURN_NO_PROMETHEUS` is defined when prometheus libraries are not present and any prometheus functionality must be disabled

While all above is correct, it does not require ifdef-ing out all related code.
For example, prometheus related fields in turn_params do not need to be compiled out. Same for certain function parameters.

This PR reduces amount of places in code where `TURN_NO_PROMETHEUS` is used to make code simpler by moving as much usage of this define into prom_server.h/c files and compiling them unconditionally.

- Always compile/link prom_server.c
- Move many TURN_NO_PROMETHEUS decisions into prom_server.c

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -31,9 +31,7 @@
 #include "mainrelay.h"
 #include "dbdrivers/dbdriver.h"
 
-#if !defined(TURN_NO_PROMETHEUS)
 #include "prom_server.h"
-#endif
 
 #if defined(WINDOWS)
     #include <Iphlpapi.h>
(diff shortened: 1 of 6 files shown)


src/apps/relay/mainrelay.c
```diff
@@ -2622,19 +2615,7 @@ int main(int argc, char **argv)
...
-	}
-	else if (prometheus_status == 1) {
-	  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Prometheus collector disabled, not started.\n");
-	}
-	else {
-	  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Prometheus collector started successfully.\n");
-	}
-#endif
-
+	start_prometheus_server();
```
eakraly, line 2618: Will figure out internally if prometheus is supporter and enabled
In other cases do nothing

PR #1029, opened 2022-10-20
Backlog fifo

sgodin:
Modify SSL backlog buffer from LIFO to queue/FIFO

If data ends up in the ssl_backlog_buffer because we are waiting for a handshake to finish, then this change ensures that the data is sent out in the proper order once the handshake completes.  Previous code was sending in LIFO order.

### DIFF
--- a/src/apps/relay/ns_ioalib_engine_impl.c
+++ b/src/apps/relay/ns_ioalib_engine_impl.c
@@ -101,6 +101,8 @@ static void socket_output_handler_bev(struct bufferevent *bev, void* arg);
 static void socket_input_handler_bev(struct bufferevent *bev, void* arg);
 static void eventcb_bev(struct bufferevent *bev, short events, void *arg);
 
+static int send_ssl_backlog_buffers(ioa_socket_handle s);
+
 static int set_accept_cb(ioa_socket_handle s, accept_cb acb, void *arg);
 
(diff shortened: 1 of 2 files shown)


sgodin:
FYI - I'm doing work for wire.com on contract.  We have been adding a federated connection link between two coturn instances.  In the process we have created a few fixes that could benefit the coturn community.  This pull request the first of a few that I will be creating.

eakraly:
Thanks @sgodin !
What was the side effect of the reverse order of packets and how did you test it (pls add the explanation to PR description :) )

sgodin:
I originally produced the problem on our federation code that is not yet public.  In this code we setup a DTLS tunnel between two coturn instances.  If a client tries to send data on the tunnel before the handshake is finished the data ends up in the SSL backlog buffer.  Once the handshake completes, the SSL backlog buffer is drained and the packets added to it are sent out.  However I noticed they were being received in the reverse order that they were sent.  After inspection I realized the backlog buffer was a LIFO and not a FIFO, so I modified it, and retested it to see the packets in the expected order.   

I'm not 100% sure how this could expose itself during normal coturn use, but it would have to happen after an ssl_send, which means data towards a TLS or DTLS client.  Since initial handshakes will be done, perhaps this would only happen for mid-session TLS re-negotiations??

PR #1030, opened 2022-10-20
Log threadId to logs to aid in multi-threaded debugging.

sgodin:


src/apps/common/ns_turn_utils.c
```diff
@@ -46,6 +46,15 @@
 
 #include <signal.h>
 
+#include <unistd.h>
+#include <sys/syscall.h>
+
+#ifndef SYS_gettid
+#error "SYS_gettid unavailable on this system"
```
eakraly, line 53: We do not want compilation to break here
Can we have default implementation?

sgodin:
@eakraly Good idea.  I'll just revert to not logging the threadid if the API is NOT available.

KangLin:
Suggest to use thirty log library. ag: log4cxx

eakraly:
> Suggest to use thirty log library. ag: log4cxx
Definitely. But that is a bigger project.

src/apps/common/ns_turn_utils.c
sgodin: @eakraly Sorry it took a few days, been a busy week.  I've modified the PR so that if SYS_gettid doesn't exist then it will still compile, and logs will be same as before this commit (ie: no TID in logs).  Thanks for all your efforts in getting my PR's merged.  :)

src/apps/common/ns_turn_utils.c
```diff
@@ -46,6 +46,14 @@
 
 #include <signal.h>
 
+#include <unistd.h>
```
eakraly, line 49: Now need to wrap it with `#if !defined(WINDOWS)` or something similar

sgodin: Done!  I had reorganize the logging part a little, due to changes on main from other pull request merges.

eakraly (approved):

PR #1031, opened 2022-10-20
ignore raw UDP if no_udp is enabled

sgodin:


eakraly:
Hi @sgodin ! Thank you for your contribution
Could you please provide some description on how this works, and how you tested it

Also, it seems to address #33

fattahi:
I checked 4.5.2 4.6.0 4.6.1 version with no-udp no-tcp and no-tls option to force stream to dtls mode. but it is not working with 4.6.2. 
I think this fix cause the problem for DTLS only.

fattahi:
> @eakraly Wow there is a LOT of detail in #33, I skimmed it. It seems to solve the problem for no_udp, but not for no_tcp. I'm sure the fix for no_tcp would be similar.
> 
> Essentially, for a DTLS client (that we haven't heard from before), the code in handle_udp_packet will have created the chs/ioa_socket in the block just above my change (see dtls_server_input_handler's call to dtls_accept_client_connection that calls create_ioa_socket_from_ssl). This only happens if the first message received from a client is a DTLS handshake. Otherwise, we have received UDP data from a new endpoint that is not a DTLS handshake, so it is raw UDP and the code just below my if statement will have created a UDP_SOCKET in the create_ioa_socket_from_fd call, allowing further processing of the RAW UDP.
> 
> This was tested by trying to perform a TURN allocation via UDP (not DTLS) when no-udp setting was enabled.

you means that my traffic in coturn before 4.6.2 is insecure and not DTLS?
I'm using CoTURN for matrix synapse home server and Element web or android as client...
I See dtls handshaking on starting voice call and alert on call termination:
`# tshark -Y "dtls"
14092 12.737988826 a.a.a.a → b.b.b.b DTLS 199 Client Hello
14180 12.806858502 b.b.b.b → a.a.a.a DTLSv1.2 659 Server Hello, Certificate, Server Key Exchange, Certificate Request, Server Hello Done
14232 12.843743802 a.a.a.a → b.b.b.b DTLSv1.2 587 Certificate, Client Key Exchange, Certificate Verify, Change Cipher Spec, Encrypted Handshake Message
14306 12.904583130 b.b.b.b → a.a.a.a DTLSv1.2 612 New Session Ticket, Change Cipher Spec, Encrypted Handshake Message
40133 32.363867570 b.b.b.b → a.a.a.a DTLSv1.2 81 Encrypted Alert
40176 32.393960425 a.a.a.a → b.b.b.b DTLSv1.2 81 Encrypted Alert`
a.a.a.a is client and b.b.b.b is ip of turn server

fattahi:
my problem is not corresponding this commit. I compared source code of 4.6.1-r3 and 4.6.2-r5 both patched for "Disallow raw UDP if no_udp is enabled". but I have problem with DTLS on 4.6.2 ...

PR #1033, opened 2022-10-20
fix bug in calls to ssl_read and ssl_send where extra verbose flag goes missing

sgodin:
-extra verbose has int value of 2, ensure it's not lost when calling ssl_read and ssl_send

### DIFF
--- a/src/apps/relay/ns_ioalib_engine_impl.c
+++ b/src/apps/relay/ns_ioalib_engine_impl.c
@@ -2584,7 +2584,7 @@ static int socket_input_worker(ioa_socket_handle s)
 		if(s->ssl && (len>0)) { /* DTLS */
 			send_ssl_backlog_buffers(s);
 			buf_elem->buf.len = (size_t)len;
-			ret = ssl_read(s->fd, s->ssl, (ioa_network_buffer_handle)buf_elem, (s->e ? s->e->verbose : 0));
+			ret = ssl_read(s->fd, s->ssl, (ioa_network_buffer_handle)buf_elem, (s->e ? s->e->verbose : TURN_VERBOSE_NONE));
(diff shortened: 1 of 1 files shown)


src/apps/relay/ns_ioalib_engine_impl.c
```diff
@@ -2584,7 +2584,7 @@ static int socket_input_worker(ioa_socket_handle s)
 		if(s->ssl && (len>0)) { /* DTLS */
 			send_ssl_backlog_buffers(s);
 			buf_elem->buf.len = (size_t)len;
-			ret = ssl_read(s->fd, s->ssl, (ioa_network_buffer_handle)buf_elem, ((s->e) && s->e->verbose));
+			ret = ssl_read(s->fd, s->ssl, (ioa_network_buffer_handle)buf_elem, (s->e ? s->e->verbose : 0));
```
eakraly, line 2587: Let's replace 0 with explicit `TURN_VERBOSE_NONE`

src/apps/relay/ns_ioalib_engine_impl.c
```diff
@@ -3279,7 +3279,7 @@ int send_data_from_ioa_socket_nbh(ioa_socket_handle s, ioa_addr* dest_addr,
 								s,
 								(char*) ioa_network_buffer_data(nbh),
 								ioa_network_buffer_get_size(nbh),
-								((s->e) && s->e->verbose));
+								(s->e ? s->e->verbose : 0));
```
eakraly, line 3282: Same as above

eakraly:
Thanks @sgodin ! This looks great!

src/apps/relay/ns_ioalib_engine_impl.c
sgodin: will do :)

PR #1042, opened 2022-10-24
Fix variable argument handling

gjasny:


src/apps/relay/libtelnet.c
```diff
@@ -1483,12 +1483,12 @@ int telnet_vprintf(telnet_t *telnet, const char *fmt, va_list va) {
 		if (output == 0) {
 			_error(telnet, __LINE__, __func__, TELNET_ENOMEM, 0,
 					"malloc() failed: %s", strerror(errno));
+			va_end(va2);
 			return -1;
 		}
 		rs = vsnprintf(output, rs + 1, fmt, va2);
 	}
 	va_end(va2);
-	va_end(va);
```
gjasny, line 1491: Wasn't `va_end` called twice: one time in the caller and one time in the callee?

ggarber:
Looks like a good catch! Thank you!

PR #1047, opened 2022-10-25
Use khash 0.2.8

gjasny:
The previously used version of khash trigged some static code analysis warnings that are gone with the latest version:

https://github.com/attractivechaos/klib/blob/9a063b33efd841fcc42d4b9f68cb78bb528bf75b/khash.h

### DIFF
--- a/src/server/ns_turn_khash.h
+++ b/src/server/ns_turn_khash.h
@@ -1,6 +1,6 @@
 /* The MIT License
 
-   Copyright (c) 2008, by Attractive Chaos <attractivechaos@aol.co.uk>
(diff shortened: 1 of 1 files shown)


eakraly:
Hi @gjasny !

Could you please specify (some) of the warnings static analysis highlighted in the commit message?
Also, any idea if any of memory/performance impacts should be expected (to your knowledge)
Just trying to make sure every commit is sufficiently documented for future reference.

Thank you!

KangLin (approved):
It is correct to use the new version.

eakraly (approved):

PR #1048, opened 2022-10-25
Fix resource leaks

gjasny:


eakraly:
Thank you for your contribution @gjasny !
Could you please add short explanation to how you found it and how you verified/tested it?
This is to provide some more info in commit history directly in git (vs in GH history)
Thank you!

KangLin:
@gjasny  Very good!
@eakraly Can merge.

ggarber (approved):

PR #1054, opened 2022-10-29
Update to fix duplicate stdout log output

kadoshita:
Disables default log output to stdout if stdout is set as logfile.

When execute `turnutils_stunclient` , the output will be duplicated as in the following example.

```
$ ./turnutils_stunclient stun.example.com
0: : IPv4. UDP reflexive addr: 203.0.113.1:10000
0: : IPv4. UDP reflexive addr: 203.0.113.1:10000
```

This is because stdout is set in `set_logfile` and furthermore log output to stdout is done by default.

In this change, call `set_no_stdout_log` to disable default log output to stdout when stdout is set as logfile.

This solves the problem of duplicate output to stdout of the same log.

### DIFF
--- a/src/apps/natdiscovery/natdiscovery.c
+++ b/src/apps/natdiscovery/natdiscovery.c
@@ -634,6 +634,7 @@ int main(int argc, char **argv)
 	if (socket_init()) return -1;
 
 	set_logfile("stdout");
+	set_no_stdout_log(1);
 	set_system_parameters(0);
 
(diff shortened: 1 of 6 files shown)


eakraly:
Thanks a lot @kadoshita !
Until now I did not think about it but now I see how annoying it was all the time....

kadoshita:
@eakraly You're welcome!

PR #1058, opened 2022-10-30
Simplify defines for macOS platform

eakraly:
`__APPLE__` is enough to detect macOS

### DIFF
--- a/src/apps/common/apputils.c
+++ b/src/apps/common/apputils.c
@@ -35,8 +35,7 @@
 
 #include <event2/event.h>
 
-#if defined(__unix__) || defined(unix) || defined(__APPLE__) \
-	|| defined(__DARWIN__) || defined(__MACH__)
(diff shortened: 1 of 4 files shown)


SamuelMarks:
@eakraly Based on this https://sourceforge.net/p/predef/wiki/OperatingSystems/#macos it's `__APPLE__ && __MACH__` to determine macOS.

Official docs here: https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html#//apple_ref/doc/uid/TP40002850-SW13

There is also `#if TARGET_OS_MAC` to be aware of if you want to discern between say, iOS and macOS.

eakraly:
Absolutely correct. In fact, before that we had a much broader filter (because of ||) so now it is a little more narrow (only __APPLE__). This "allows" building for other Apple platforms (not macOS) but that is not something easily possible out of box with our current build systems support.

Issue #1044, opened 2022-10-25
Requires a supported operating system and compiler


KangLin:
Requires a supported operating system and compiler.

## OS
  - [x] Windows: Min version windows7
  - [x] Apple
  - Unix
  - Linux
    - [x] Ubuntu
    - [x] debian

## Compiler
    - [x] MSVC: Min version MSVC 2015
    - [x] MINGW: 
    - [x] GCC

## Macro
  -  WINDOWS: Support the all compilers of windows(ag. MSVC, MinGW, CGWin etc)
  - _MSC_VER: Only support MSVC compiler
  - MINGW: Only support MinGW compiler
  - __CYGWIN__:
  - __CYGWIN32__:
  - __CYGWIN64__:
  - LINUX
  - UNIX
  - __APPLE__

Please add the operating systems and compilations you can support below.

KangLin:
The following macros are not required: Win32, Win64, __CYGWIN32__, __CYGWIN64__, MINGW32, MINGW64

Neustradamus:
WINE?
ReactOS?

KangLin:
> WINE? ReactOS?

Ok! It has been added, and I hope someone will support them.

PR #1061, opened 2022-10-31
Fix socket handling on Windows

pando-emil:
* Make non-blocking calls non-blocking on windows as well
* Wrap errno-handling in inline functions

### DIFF
--- a/src/apps/natdiscovery/natdiscovery.c
+++ b/src/apps/natdiscovery/natdiscovery.c
@@ -139,7 +139,7 @@ static int stunclient_send(int sockfd, ioa_addr *local_addr, int *local_port, io
 
 		do {
 			len = sendto(sockfd, req.getRawBuffer(), req.getSize(), 0, (struct sockaddr*) remote_addr, (socklen_t) slen);
-		} while (len < 0 && ((errno == EINTR) || (errno == ENOBUFS) || (errno == EAGAIN)));
+		} while (len < 0 && (socket_eintr() || socket_enobufs() || socket_eagain()));
 
 		if (len < 0)
 		err(-1, NULL);
@@ -180,9 +180,9 @@ static int stunclient_receive(int sockfd, ioa_addr *local_addr, ioa_addr *reflex
 				ptr += len;
 				break;
 			}
-		} while (len < 0 && (errno == EINTR));
+		} while (len < 0 && socket_eintr());
 
-		if ((errno == EAGAIN) || (errno == EWOULDBLOCK)) {
+		if (socket_eagain() || socket_ewouldblock()) {
 			printf("STUN receive timeout..\n");
 			ret = 1;
 			return ret;
@@ -274,7 +274,7 @@ static int run_stunclient(ioa_addr *local_addr, ioa_addr *remote_addr, ioa_addr 
 	ret=init_socket(&udp_fd, local_addr, *local_port, remote_addr);
 	ret=stunclient_send(udp_fd, local_addr, local_port, remote_addr, change_ip, change_port, padding, -1);
 	ret=stunclient_receive(udp_fd, local_addr, reflexive_addr, other_addr, rfc5780);
-	close(udp_fd);
+	socket_closesocket(udp_fd);
 
 	return ret;
 }
@@ -300,8 +300,8 @@ static int run_stunclient_hairpinning(ioa_addr *local_addr, ioa_addr *remote_add
 	if(ret){
 		ret=stunclient_receive(udp_fd2, local_addr, reflexive_addr, other_addr, rfc5780);
 	}
-	close(udp_fd);
-	close(udp_fd2);
+	socket_closesocket(udp_fd);
+	socket_closesocket(udp_fd2);
 
 	return ret;
 }
@@ -374,7 +374,7 @@ static int stunclient_send(stun_buffer *buf, int sockfd, ioa_addr *local_addr, i
 
 		do {
 			len = sendto(sockfd, buf->buf, buf->len, 0, (struct sockaddr*) remote_addr, (socklen_t) slen);
-		} while (len < 0 && ((errno == EINTR) || (errno == ENOBUFS) || (errno == EAGAIN)));
+		} while (len < 0 && (socket_eintr() || socket_enobufs() || socket_eagain()));
 
 		if (len < 0)
 		err(-1, NULL);
@@ -414,7 +414,7 @@ static int stunclient_receive(stun_buffer *buf, int sockfd, ioa_addr *local_addr
 				ptr += len;
 				break;
 			}
-		} while (len < 0 && (errno == EINTR));
+		} while (len < 0 && socket_eintr());
 
 		if (recvd > 0)
 		len = recvd;
@@ -482,7 +482,7 @@ static int stunclient_receive(stun_buffer *buf, int sockfd, ioa_addr *local_addr
 				ret=1;
 			}
 		} else {
-			if ((errno == EAGAIN) || (errno == EWOULDBLOCK)) {
+			if (socket_eagain() || socket_ewouldblock()) {
 				printf("STUN receive timeout..\n");
 			}else{
 				printf("The response is not a STUN message\n");
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,23 @@ IF(MSVC)
     add_definitions(-D_CRT_SECURE_NO_WARNINGS)
 ENDIF(MSVC)
 
+if(WIN32)
+    # We have to set _WIN32_WINNT
+    if(${CMAKE_SYSTEM_VERSION} EQUAL 10) # Windows 10
+        add_definitions(-D _WIN32_WINNT=0x0A00)
+    elseif(${CMAKE_SYSTEM_VERSION} EQUAL 6.3) # Windows 8.1
+        add_definitions(-D _WIN32_WINNT=0x0603)
+    elseif(${CMAKE_SYSTEM_VERSION} EQUAL 6.2) # Windows 8
+        add_definitions(-D _WIN32_WINNT=0x0602)
+    elseif(${CMAKE_SYSTEM_VERSION} EQUAL 6.1) # Windows 7
+        add_definitions(-D _WIN32_WINNT=0x0601)
+    elseif(${CMAKE_SYSTEM_VERSION} EQUAL 6.0) # Windows Vista
+        add_definitions(-D _WIN32_WINNT=0x0600)
+    else() # Windows XP (5.1)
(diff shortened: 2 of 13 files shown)


src/apps/natdiscovery/natdiscovery.c
```diff
@@ -300,8 +300,8 @@ static int run_stunclient_hairpinning(ioa_addr *local_addr, ioa_addr *remote_add
 	if(ret){
 		ret=stunclient_receive(udp_fd2, local_addr, reflexive_addr, other_addr, rfc5780);
 	}
-	close(udp_fd);
-	close(udp_fd2);
+	socket_closesocket(udp_fd);
```
eakraly, line 303: This is also somehow gets converted to plain `close()` using macro?
I feel this is an unexpected and undesired side-effect

pando-emil: Yeah, the macro is defined src/apps/common/apputils.h and has been since the beginning of time (well, since the code was imported to git).
While both unix socket and winsock opens a socket with socket, winsock closes a socket with closesocket. In what way do you mean it is an undesired side-effect to use socket_closesocket()?

eakraly: Sorry I thought the new macros does that. Disregard my comment then.

eakraly:
@KangLin you should probably take a look as well. TY!

src/apps/natdiscovery/natdiscovery.c
pando-emil: Ahh, good, then I'm less confused :)
I am aware those expansion macros are kind of a brainfuck to read. But all they do is to create inline functions for checking errno (or WSAGetLastError() in case of Windows) against each error error code.

The fact that errno and for example close() still exists when you build on Windows, but don't apply for sockets, makes things a bit complicated. There is always the risk that a new contributor will add new calls to close() and new checks using errno directly. Everything will compile fine, but not behaving as expected on Windows.

eakraly (approved):

eakraly:
Hi @pando-emil ! Thanks for the change!
Could you please explain how you tested and what results did you observe?
Thanks a lot!

pando-emil:
Hi @eakraly 
Our use case is with the TURN server, so that is what we have been testing most. Our typical test setup opens 12 video+audio streams in parallel, and let them run for an extensive period.
Before the changes in this PR, one stream was working kind of okay, but with a bunch of false "read returned error 0"-printouts due to the errno handling. When running with more than one stream the streams went very bursty due to the blocking sockets in dtls_listener.c.
After the changes in this PR multiple streams runs smoothly in our setup.
However I have to be honest and admit that we haven't done much of regression testing on other platforms...

KangLin:
@pando-emil Can you provide test data？

pando-emil:
Hi @KangLin
What kind of test data do you have in mind? Test data as in a network capture that shows the result, or test data as in instructions on how to reproduce the tests?

KangLin:
@pando-emil Test data as in a network capture that shows the result.
Please test in Linux, wether has same problem?

在 2022-11-02 21:26:44，"Emil Ljungdahl" ***@***.***> 写道：


Hi @KangLin
What kind of test data do you have in mind? Test data as in a network capture that shows the result, or test data as in instructions on how to reproduce the tests?

pando-emil:
Hi @KangLin 
Okay, it will take some time to generate captures, I won't be at the office until next week.
Why would the linux-build have the same issue!? The blocking sockets are only present within windows ifdefs, you should know, you added the TODOs I implemented :)

KangLin:
>  Why would the linux-build have the same issue!? The blocking sockets are only present within windows ifdefs, you should know, you added the TODOs I implemented :)

Oh, I misunderstood what you meant earlier.

Also, my review problem, please solve it.
Thank you!

pando-emil:
Ahh, okay, good! :)

> Also, my review problem, please solve it.
Do you mean you left a review note? In that case, I don't think I can see it?

KangLin:
> Do you mean you left a review note? In that case, I don't think I can see it?

Yes. I am left note under code. I've had similar display issues before, and I've thought that only me will appear. You can try refreshing the page a few more times.

pando-emil:
Nope, sorry, the only comment I can see is from eakraly 3 days ago?

KangLin:
@pando-emil  You try to `Files Changed` lable.
@eakraly Can you see my review note?

KangLin:
- _WIN32_WINNT
  - Do not define macros with the prefix '_', which are reserved by the compiler.
  - Do not define and use macros of the form WIN32_WININT, please use WINDOWS_VERSION.
  - The macro you define is not used in the code. so don't need it?
  - See #1044 
- EPERM don't need to define socket_eperm()?
  You should forget it.
- ERRNO_COMPARE_FUN(enomem, WSA_NOT_ENOUGH_MEMORY)
It is recommended to define inline functions, use macros, and the readability becomes deterioration.

pando-emil:
I did look at the Files changed label, but can't see anything there. Anyhow, please see my comments below:

- _WIN32_WINNT 
It is very intentionally to set the _WIN32_WINNT, it is used inside winsock2.h and affects which socket APIs that are available. More info: https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt

- EPERM don't need to define socket_eperm()?
There is no equivalent error code on Windows, and since SO_BINDTODEVICE is not present on Windows I don't think it will be used. But I agree, at least I could have added a comment about it.

- ERRNO_COMPARE_FUN(enomem, WSA_NOT_ENOUGH_MEMORY)
That macro actually expands to an inline-function! Initially I did write out all the inline functions, but in my opinion the code became even harder to read by doing that.

KangLin:
> It is very intentionally to set the _WIN32_WINNT, it is used inside winsock2.h and affects which socket APIs that are available. More info: https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt

Yes. but it is not currently being made.

> * That macro actually expands to an inline-function! Initially I did write out all the inline functions, but in my opinion the code became even harder to read by doing that.

The function declaration is placed in the header file, and the function body is placed in the .cpp file, so that it is not messy.

- case EACCES:  Did you forget this section?

pando-emil:
>> It is very intentionally to set the _WIN32_WINNT, it is used inside winsock2.h and affects which socket APIs that are available. More info: https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt

> Yes. but it is not currently being made.

What do you mean?
At least with mingw when cross compiling, _WIN32_WINNT will default to 0x0501 (a.k.a. Windows XP). This will give compile warnings on for example inet_ntop (which is used in coturn).

>> That macro actually expands to an inline-function! Initially I did write out all the inline functions, but in my opinion the code became even harder to read by doing that.

> The function declaration is placed in the header file, and the function body is placed in the .cpp file, so that it is not messy.

My experience with inline functions implemented in .c-files is that they rarely get inlined properly by the compiler when used from other c-files. But that might just be me getting old and grumpy after using too many obscure compilers...

> case EACCES: Did you forget this section?

Good catch! Fix comitted

KangLin:
> What do you mean? At least with mingw when cross compiling, _WIN32_WINNT will default to 0x0501 (a.k.a. Windows XP). This will give compile warnings on for example inet_ntop (which is used in coturn).

I am think to the macro is defined by compiler. but I amn't sure that it  is defined in mingw. wait that I am test it tomorrow.


> My experience with inline functions implemented in .c-files is that they rarely get inlined properly by the compiler when used from other c-files. But that might just be me getting old and grumpy after using too many obscure compilers...

Haha, I also used such compilers a long long ago, but they were before c99, or compilers that did not comply with c standards. But now we are clearly used c11. Second, use macros to define functions, which are not recognized by IntelliSense in some lightweight IDEs, so it is highly recommended to define functions directly.

在 2022-11-03 19:42:39，"Emil Ljungdahl" ***@***.***> 写道：


It is very intentionally to set the _WIN32_WINNT, it is used inside winsock2.h and affects which socket APIs that are available. More info: https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt

Yes. but it is not currently being made.

What do you mean?
At least with mingw when cross compiling, _WIN32_WINNT will default to 0x0501 (a.k.a. Windows XP). This will give compile warnings on for example inet_ntop (which is used in coturn).

That macro actually expands to an inline-function! Initially I did write out all the inline functions, but in my opinion the code became even harder to read by doing that.

The function declaration is placed in the header file, and the function body is placed in the .cpp file, so that it is not messy.

My experience with inline functions implemented in .c-files is that they rarely get inlined properly by the compiler when used from other c-files. But that might just be me getting old and grumpy after using too many obscure compilers...

case EACCES: Did you forget this section?

Good catch! Fix comitted

KangLin:
These functions are not used very much when running, I think it is enough to use normal functions, and there is no need to use inline functions.

KangLin:
@pando-emil 
- In MSVC2017, The _WIN32_WINNT is defined in sdkddkver.h

```c
#if !defined(_WIN32_WINNT) && !defined(_CHICAGO_)
#define  _WIN32_WINNT   0x0A00
#endif
```

- MinGW, The _WIN32_WINNT is defined in _mingw.h:

```c
#define _WIN32_WINNT 0x601
```

My Msys and Mingw version:

```bash
$ uname -a
MINGW32_NT-10.0-19044 DESKTOP-081RDD4 3.3.6-341.x86_64 2022-09-06 08:02 UTC x86_64 Msys

$ gcc --version
gcc.exe (Rev2, Built by MSYS2 project) 12.1.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

```

I am compile coturn, there are no warnings you said.
**The macros with the prefix '_', which are reserved by the compiler.** so we cann't define it.

eakraly:
Hi @pando-emil !
As usual I would suggest to split this PR into 2 parts:
- 1 is trivial - Wrap errno-handling in inline functions: this is very straightforward change and can be submitted quickly
- 2 is the async change - it is actually small but seemingly need thorough review

WDYT?

pando-emil:
Hi @eakraly 

Yep, will do that! Actually I think I make that 3 different PR. The _WIN32_WINNT part also stands for itself, and there is obviously room for debate :)

Neustradamus:
To follow

eakraly:
Hi @pando-emil ! Looking forward to progression on this PR :)

pando-emil:
Hi @eakraly 

Yeah I know, sorry for the delay! I'll try to find time for this asap.

Issue #1045, opened 2022-10-25
Proper type in sizeof used?


gjasny:
Hello,

our static code analysis suggested to take a deeper look at the following two sizeof expressions:

https://github.com/coturn/coturn/blob/cfa5f66cd7ad9a085ee32ffdadafcc84e3093750/src/apps/relay/ns_ioalib_engine_impl.c#L3882

https://github.com/coturn/coturn/blob/cfa5f66cd7ad9a085ee32ffdadafcc84e3093750/src/apps/relay/ns_ioalib_engine_impl.c#L3952

Maybe `sizeof(size_t*)` should be changed into `sizeof(size_t)`. Could you please have a look?

Thanks,
Gregor

eakraly:
Great finding @gjasny 
I agree that it should be `sizeof(size_t)`.
We are lucky that most (or all) of the time `sizeof(size_t)` == `sizeof(size_t*)`

Will you please submit a PR and verify with your static code analysis tool?
Thank you!

PR #1063, opened 2022-10-31
Properly calculate size for sm_allocated

gjasny:
The `sm_allocated` field is an array of `size_t` values. Therefore its size must be calculated as `n * sizeof(size_t)`.

Our static code analysis tool complained:
```
Sizeof not portable (SIZEOF_MISMATCH)
suspicious_sizeof: Passing argument r->sm_allocated of type size_t * and argument (r->sm_chunk + 1) * 8UL /* sizeof (size_t *) */ to function realloc is suspicious. In this case, sizeof (size_t *) is equal to sizeof (size_t), but this is not a portable assumption.
```

Fix: #1045


eakraly:
Thank you

eakraly (approved):

PR #1065, opened 2022-10-31
Do not discard qualifiers in free()

gjasny:
Warning emitted from GCC:
```
warning: passing argument 1 of ‘free’ discards ‘const’ qualifier from pointer target type
```


eakraly:
Thank you @gjasny

PR #1075, opened 2022-11-03
feature(1026): add sessioncount to prometheus metrics

paulkram:


src/apps/relay/prom_server.h
```diff
@@ -59,8 +59,8 @@ void start_prometheus_server(void);
 
 void prom_set_finished_traffic(const char* realm, const char* user, unsigned long rsvp, unsigned long rsvb, unsigned long sentp, unsigned long sentb, bool peer);
 
-void prom_inc_allocation(void);
-void prom_dec_allocation(void);
+void prom_inc_allocation(SOCKET_TYPE type);
+void prom_dec_allocation(SOCKET_TYPE type);
 #else
 
 void start_prometheus_server(void);
```
paulkram, line 66: I think this method does not need to be guarded by the `TURN_NO_PROMETHEUS ` define.

eakraly: I agree - we should have minimal amount if ifdefs outside of prom_server.c file to improve readability of the code

src/apps/relay/prom_server.c
```diff
@@ -129,15 +129,17 @@ void prom_set_finished_traffic(const char* realm, const char* user, unsigned lon
   }
 }
 
-void prom_inc_allocation(void) {
+void prom_inc_allocation(SOCKET_TYPE type) {
   if (turn_params.prometheus == 1){
-    prom_gauge_inc(turn_total_allocations, NULL);
+    prom_gauge_inc(turn_total_allocations, (const char*[]) {"all"});
```
eakraly, line 134: Not required
We only need to reports the type and the aggregation will be done in prometheus query. For example, this explicit query for all types
```
turn_total_allocations{type~=".*"}
```
If we report both "all" and "udp" then every allocation will be counted twice (in that query)

src/apps/relay/prom_server.c
```diff
@@ -129,15 +129,17 @@ void prom_set_finished_traffic(const char* realm, const char* user, unsigned lon
...
+    prom_gauge_inc(turn_total_allocations, (const char*[]) {socket_type_name(type)});
   }
 }
 
-void prom_dec_allocation(void) {
+void prom_dec_allocation(SOCKET_TYPE type) {
   if (turn_params.prometheus == 1){
-    prom_gauge_dec(turn_total_allocations, NULL);
+    prom_gauge_dec(turn_total_allocations, (const char*[]) {"all"});
```
eakraly, line 141: Same as above

src/server/ns_turn_maps.c
```diff
@@ -201,7 +201,7 @@ int ur_map_foreach(ur_map* map, foreachcb_type func) {
   return 0;
 }
 
-int ur_map_foreach_arg(ur_map* map, foreachcb_arg_type func, void* arg) {
+int ur_map_foreach_arg(const ur_map* map, foreachcb_arg_type func, void* arg) {
```
eakraly, line 204: Do we want this in this PR? What does it fix?

eakraly:
Hi @paulkram !
I checked out your PR and found one more issue: I always get socket type UDP in prometheus metric - not sure if this is me testing it incorrectly or it is broken somehow (I tried with -t -T)
Also, as I suspected, "ALL" and "UDP" are duplicates

Thanks!

src/server/ns_turn_maps.c
paulkram: I had another approach how to add more metrics at first and missed to remove that. This only improves readability for new devs to understand which are in and which are out parameters.

src/apps/relay/prom_server.c
paulkram: Actually, it's a different number. "all" should count all independent of the type and "udp" should be only udp. During my tests I saw at least "udp" and "tcp". Still, I totally agree that prometheus or the tool to draw graphs can then be used to calculate "all". If you prefer, I can remove the "all" field again.

paulkram:
Hi Pavel,
I saw at least "udp" and "tcp" during my tests but I ran it with some clients and more in a chaotic scenario. We should monitor the values (I'll do) and if there is something broken, I'm going to open another PR.

 

 
 

Gesendet: Sonntag, 06. November 2022 um 19:16 Uhr
Von: "Pavel Punsky" ***@***.***>
An: "coturn/coturn" ***@***.***>
Cc: "Paul Kramer" ***@***.***>, "Mention" ***@***.***>
Betreff: Re: [coturn/coturn] feature(1026): add sessioncount to prometheus metrics (PR #1075)


 

Hi @paulkram !
I checked out your PR and found one more issue: I always get socket type UDP in prometheus metric - not sure if this is me testing it incorrectly or it is broken somehow (I tried with -t -T)
Also, as I suspected, "ALL" and "UDP" are duplicates

Thanks!

PR #1087, opened 2022-11-13
Update Changelog and Readme

eakraly:
Authors generated using:
```
git log --pretty=format:"%an" 4.6.0..HEAD|sort|uniq
```

Commit log:
```
git log --pretty=format:"- %s (%an)" 4.6.0..HEAD
```

Readme.md files updated by rearranging existing file (and loosely following format suggested by https://github.com/jehna/readme-best-practices)


eakraly:
Not sure how the changelog was generated before - this format is easy to generate
Please review if makes sense @ggarber

ChangeLog
```diff
@@ -1,3 +1,93 @@
...
+	- Samuel Marks (807580+SamuelMarks@users.noreply.github.com)
+	- Scott Godin (sgodin@sipspectrum.com)
+	- Thibaut ACKERMANN (thib-ack@users.noreply.github.com)
+	- Tom Bevan (thehorrorthehorror@gmail.com)
+	- Yoshiki Kadoshita (sublimer.me@gmail.com)
+	- tyranron (tyranron@gmail.com)
+
+Changelist:
+	- Do not build debug version during PR preflight (Pavel Punsky)
+	- Bump actions/cache from 2 to 3 (#1084) (dependabot[bot])
```
tyranron, line 22: @eakraly I think we may remove such lines from changelog after auto-generating as they make not much sense for readers and only pollute the list of changes. Changelog, imho, diverges from git log in the way that it should contain only changes visible for the end-users.

tyranron (changes requested):

ChangeLog
eakraly: True
I wanted to delete it but decided to keep it for review just in case

ChangeLog
```diff
@@ -1,3 +1,90 @@
...
+	- Add CI tests in older ubuntu version (#981) (Gustavo Garcia)
+	- Replace bzero with memset (#986) (Pavel Punsky)
+	- Fix a warning (#988) (Pavel Punsky)
+	- Improve openssl3 and FIPS support (#955) (Pavel Punsky)
+	- Preserve file timestamps when using install(1) (#983) (Robert Scheck)
+	- Fix renegotiation flag for older version of openssl (#978) (Pavel Punsky)
+	- Improve GitHub release notes of Docker image (tyranron)
+	- Improve Docker CI pipeline (tyranron)
+	- Upgrade Docker image to 4.6.0 Coturn version (tyranron)
+	- Fix ChangeLog for 4.6.0 release (#976) (tyranron)
```
tyranron, line 86: @eakraly changes regarding Docker image may be removed too. We track them in a separate changelog. Here is better to keep changes regarding Coturn itself.

ChangeLog
```diff
@@ -1,3 +1,90 @@
...
+	- Tom Bevan (thehorrorthehorror@gmail.com)
+	- Yoshiki Kadoshita (sublimer.me@gmail.com)
+	- tyranron (tyranron@gmail.com)
+
+Changelist:
+	- Do not build debug version during PR preflight (Pavel Punsky)
+	- Add clang format rules and checks (#935) (Gustavo Garcia)
+	- Fix duplicate prometheus metric report (#1079) (Pavel Punsky)
+	- feature(1026): add sessioncount to prometheus metrics (#1075) (Paul Kramer)
+	- Add macOS build action (#1059) (Pavel Punsky)
```
tyranron, line 25: Seems to be toolchain-related only.

tyranron (changes requested):

Issue #1049, opened 2022-10-27
Update old files


Neustradamus:
Dear @coturn team,

Can you update:
- https://github.com/coturn/coturn/blob/master/AUTHORS
- https://github.com/coturn/coturn/blob/master/README.md
- https://github.com/coturn/coturn/blob/master/README.turnadmin
- https://github.com/coturn/coturn/blob/master/README.turnserver
- https://github.com/coturn/coturn/blob/master/README.turnserver
- https://github.com/coturn/coturn/blob/master/STATUS
- https://github.com/coturn/coturn/blob/master/TODO
- https://github.com/coturn/coturn/blob/master/LICENSE.OpenSSL (here: https://github.com/openssl/openssl/blob/master/LICENSE.txt)
- https://github.com/coturn/coturn/blob/master/LICENSE

The coturn/CoTURN/COTURN term is missing too.

Maybe some files can be removed.

Maybe I have forgotten some files too.

eakraly:
Thanks @Neustradamus !
This is definitely an important task. Our internal plan is to have some groundwork/maintenance done first and then get to the documentation sometime EOY 2022 or early next year 🤞.

eakraly:
- [ ] https://github.com/coturn/coturn/blob/master/AUTHORS
- [x] https://github.com/coturn/coturn/blob/master/README.md
- [ ] https://github.com/coturn/coturn/blob/master/README.turnadmin
- [ ] https://github.com/coturn/coturn/blob/master/README.turnserver
- [ ] https://github.com/coturn/coturn/blob/master/README.turnserver
- [ ] https://github.com/coturn/coturn/blob/master/STATUS
- [ ] https://github.com/coturn/coturn/blob/master/TODO
- [ ] https://github.com/coturn/coturn/blob/master/LICENSE.OpenSSL (here: https://github.com/openssl/openssl/blob/master/LICENSE.txt)
- [ ] https://github.com/coturn/coturn/blob/master/LICENSE

eakraly:
@Neustradamus @ggarber 
Regarding licences:
1. What is the significance of including openssl license? There are many more libraries that are used by coturn and their licenses are not referenced or committed (libevent at the very least). To my knowledge we can omit it (remove the file)

2. https://github.com/coturn/coturn/blob/master/LICENSE is the license of the project: TBH not sure what we can/should do about it.

Ideas/suggestions?

Neustradamus:
There is OpenSSL code here:
[...]

PR #1093, opened 2022-11-17
Delete AUTHORS

eakraly:
Deleting the file as it does not represent (and it is very hard to do so in the format it has right now) all the contributors to the project I suggest to signify the contribution of the original author (Oleg Moskalenko @mom040267) in a different form.

#1049


KangLin:
This document is required for licensing. The file cannot be deleted. But it can be modified.

PR #1096, opened 2022-11-17
Move and split documentation files

eakraly:
This is not a long term documentation works - this is a temp proposal to modernize existing documentation

- Create docs folder
- Move existing documentation files into docs
- Split one massive INSTALL files into smaller files (no text editing was done except small markdown header formatting)
- Rename files to be markdowns for nicer rendering
- Bad md rendering is expected in many places


eakraly:
Also, current build and package scripts embed all documentation files (in their current text form) in packages
This need to be reviewed if needed as breaks all builds

My suggestion would be to package man files but not anything else - today it is more convenient to have the documentation online
The other option is to package docs folder with whatever ends up in it

KangLin:
The help documents must be packaged.

INSTALL
```diff
@@ -1,1247 +1 @@
...
-TLS-over-SCTP.
-
-The relay side is not changing - the relay communications will still be UDP
-or TCP.
-
-XXV. Prometheus support.
-
-See for source and releases for debian packages:
-https://github.com/digitalocean/prometheus-client-c
+See https://github.com/coturn/coturn/docs
```
KangLin, line 1: Because threr are the docs folder in source code tree. so that :
```
See: [docs](docs)
```

KangLin (changes requested):

INSTALL
eakraly: Right, my concern is about people who install and now expect to open INSTALL file and see the documentation and it is not there - so there is a full link for them

But I agree - maybe we should have both
Eventually we may transfer all documentation online and only package man pages...

ggarber: Agree with KangLin here.
Let's make it "See docs folder" or something like that.

ggarber (approved):
Approved with comment

PR #1097, opened 2022-11-18
Add release.sh

KangLin:
Work on release version.
Include:
- Replease the version to new version in the version-related files
- Automation generate ChangeLog
- Automation generate AUTHORS


release.sh
```diff
@@ -0,0 +1,84 @@
...
+# Generate AUTHORS
+echo "Thanks to the following contributors (ranking will not be in order, in alphabetical order):" > ${SOURCE_DIR}/AUTHORS
+echo "" >> ${SOURCE_DIR}/AUTHORS
+echo "`git log --pretty=format:'%an <%ae>' |sort|uniq`" >> ${SOURCE_DIR}/AUTHORS
+echo "" >> ${SOURCE_DIR}/AUTHORS
+echo "Use the following command to query the author's commit:" >> ${SOURCE_DIR}/AUTHORS
+echo "   git log --author='author'" >> ${SOURCE_DIR}/AUTHORS
+echo "Or see [ChangeLog](ChangeLog)" >> ${SOURCE_DIR}/AUTHORS
+
+if [ -n "$1" ]; then
```
eakraly, line 77: This if is not necessary: $1 is mandatory (exits and prints usage if not provided)

eakraly: In general I would say this whole block is so-so as committing and pushing looks like surprising side-effect
If I run this script multiple times then I get multiple commits in remote branch
Let's say you want to change format a little - you get unnecessary pushes...

I would not include it at all

release.sh
```diff
@@ -0,0 +1,84 @@
...
+DEBIAN_VERSION=`echo ${VERSION}|cut -d "v" -f 2`
+
+# Generate ChangeLog
+echo "$MESSAGE" >>  ${SOURCE_DIR}/ChangeLog.tmp
+echo "" >> ${SOURCE_DIR}/ChangeLog.tmp
+echo "Changelist:" >> ${SOURCE_DIR}/ChangeLog.tmp
+echo "`git log --pretty=format:'- %s (%an <%ae>)' ${PRE_TAG}..HEAD`" >> ${SOURCE_DIR}/ChangeLog.tmp
+echo "" >> ${SOURCE_DIR}/ChangeLog.tmp
+echo "Contributors:" >> ${SOURCE_DIR}/ChangeLog.tmp
+echo "`git log --pretty=format:'%an <%ae>' ${PRE_TAG}..HEAD|sort|uniq`" >> ${SOURCE_DIR}/ChangeLog.tmp
```
eakraly, line 60: Should filter out commits by changebot

release.sh
```diff
@@ -0,0 +1,84 @@
...
+    VERSION=`git describe --tags`
+    if [ -z "$VERSION" ]; then
+        VERSION=`git rev-parse HEAD`
+    fi
+
+    if [ -n "$2" ]; then
+        MESSAGE=$2
+    fi
+    PRE_TAG=`git tag --sort=-taggerdate | head -n 1`
+    echo "Current verion: $VERSION, current tag: $PRE_TAG. The version to will be set: $1 $MESSAGE"
```
eakraly, line 18: Not required, just run the script

release.sh
KangLin: This script is for distribution only. If you need to modify the format and other requirements, you need to write a separate script.

release.sh
KangLin: Don't understand how the bot works, modify if needed.

release.sh
KangLin: This is required. Used to prompt you for preparation, check that you entered the correct version number.

release.sh
```diff
@@ -0,0 +1,84 @@
...
+if [ -n "$1" ]; then
+    VERSION=`git describe --tags`
+    if [ -z "$VERSION" ]; then
+        VERSION=`git rev-parse HEAD`
+    fi
+
+    if [ -n "$2" ]; then
+        MESSAGE=$2
+    fi
+    PRE_TAG=`git tag --sort=-taggerdate | head -n 1`
```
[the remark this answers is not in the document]
KangLin: Yes. I found the same issue, can you help fix it.

KangLin:
@eakraly  This one just provides an experience from my previous work. A solution for you. If you agree, you can improve it according to your project needs.

release.sh
eakraly: I think it is good - only a few fixes i would like to see
Definitely prefer script to manual work
Thank you!

eakraly:
Scripts transferred to #1101

PR #1105, opened 2022-11-24
Fix inverted logic in TLS configuration options

sysvinit:
This PR fixes some errata from #996 and #989:

- Some DTLS code was left over in the common path for allocating and initialising `SSL_CTX`'s, and the DTLS-specific configuration code was erroneously operating on the TLS context instead of the DTLS context.

- In both the TLS and DTLS codepaths, the previous refactoring inverted the logic for the `--no-tlsv1`/`--no-tlsv1_1`/etc command line options, so that these options would instead *enable* the respective (D)TLS versions, instead of disabling this. This would mean that by default coturn would only support TLS 1.3 and DTLS 1.2, and no earlier versions.

I've also regenerated the manual pages (with the `make-man.sh` script) to match the documentation in the README files.

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -3677,16 +3677,6 @@ static void set_ctx(SSL_CTX **out, const char *protocol, const SSL_METHOD *metho
     op |= SSL_OP_NO_SSLv3;
 #endif
 
-#if defined(SSL_OP_NO_DTLSv1) && DTLS_SUPPORTED
-    if (turn_params.no_tlsv1)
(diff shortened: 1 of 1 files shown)


eakraly:
Thanks @sysvinit !

We are trying to stick to "one commit - one change" approach - could you please extract man pages generation to a separate PR? Though you may want to keep TLS specific changes (if they are manual)

And thanks again!

sysvinit:
Manual page regeneration is now in #1117.

eakraly (approved):

PR #1106, opened 2022-11-24
Add explicit SIGTERM and SIGINT handlers.

sysvinit:
coturn running inside a docker container runs as PID 1, however PID 1 has special signal handling semantics (see the note at the bottom of the section [here](https://docs.docker.com/engine/reference/run/#foreground)). coturn relies on the default behaviour of SIGTERM to terminate the process, however as no signal handler is explicitly installed, it doesn't respond to SIGTERM when running inside a container. This PR fixes this problem by installing explicit signal handlers for SIGINT and SIGTERM, which trigger the same termination mechanism as the admin interface "halt" command.

This is a port of wireapp#6 for upstream.

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -265,6 +265,8 @@ static char procgroupname[1025] = "\0";
 static void read_config_file(int argc, char **argv, int pass);
 static void reload_ssl_certs(evutil_socket_t sock, short events, void *args);
 
+static void shutdown_handler(evutil_socket_t sock, short events, void *args);
+
 //////////////////////////////////////////////////
 
 static int make_local_listeners_list(void) {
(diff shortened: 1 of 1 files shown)


src/apps/relay/mainrelay.c
```diff
@@ -3831,4 +3838,13 @@ static void reload_ssl_certs(evutil_socket_t sock, short events, void *args) {
   UNUSED_ARG(args);
 }
 
+static void shutdown_handler(evutil_socket_t sock, short events, void *args) {
+  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Caught signal, terminating\n");
```
eakraly, line 3842: We cannot know what signal was sent
Should we have different handlers (per signal) instead?
Will look like code duplication but eventually will give us opportunity to handle different signals in different way (not to mention proper logging)

Issue #902, opened 2022-04-26
 ERROR: Cannot open PostgreSQL DB connection


chauhan19911:
Hi,
While connecting to the external PostgreSQL database, installed in my Windows 11 machine, through docker I'm getting this error:
`ERROR: Cannot open PostgreSQL DB connection: <host=192.168.31.173>, runtime error`

I'm using this command :
`docker run -d -p 3478:3478 -p 3478:3478/udp -p 5349:5349 -p 5349:5349/udp -p 49160-49200:49160-49200/udp coturn/coturn -n --log-file=stdout --external-ip='$(detect-external-ip)' --min-port=49160 --max-port=49200 -e, --psql-userdb="host=192.168.31.173 port=5432 dbname=webrtc connect_timeout=10"`

In C:\Program Files\PostgreSQL\14\data\pg_hba.conf, I have give permission to my IP:
`host 	all 			all 			192.168.31.173/16  		trust`

ggarber:
Doesn't look like a coturn issue.  Can you get into the docker container and try to connect to the PG server from there?

```docker exec -it CONTAINER_ID /bin/bash```

And then you can do a telnet to test basic connectivity or a psql cli.

PR #1110, opened 2022-11-28
Avoid to expand docker arguments in docker-entrypoint.sh

shu-mutou:
Not to expand arguments and to able to use `--mysql-userdb` in docker, simply use `exec "$@"` instead `exec $(eval "echo $@")`.

Fixes: #902


shu-mutou:
https://github.com/coturn/coturn/commit/d1b7acb0a52e8b6fa2fd74e4b474cbc8f0566f01 fixes only commit message, thanks in advance!

tyranron (changes requested):
@shu-mutou expanding arguments was done not just out of a random, but on purpose: allow using `$(detect-external-ip)` as a command argument (examples are given in README).

We couldn't just remove this feature. Either it should be preserved or an alternative proposed.

Regarding #902: wouldn't single quotes `'` resolve the issue like this `--min-port=49160 --max-port=49200 -e --psql-userdb='host=192.168.31.173 port=5432 dbname=webrtc connect_timeout=10'`?

shu-mutou:
To use it from shell script for CD , I also use envs in `--mysqldb-userdb` like `--mysqldb-userdb="host=${MYSQL_HOST} dbname=${MYSQL_DBNAME} user=${MYSQL_USER} password=${MYSQL_PASS} port=${MYSQL_PORT}"`.
So current implement is over expanding for me. 

What about adding envs to run `detect-external-ip` like `DETECT_EXTERNAL_IP` and `DETECT_EXTERNAL_IPV6`.

I'll investigate, so please wait for a while.

shu-mutou:
`$(eval "echo $@")` seems to expand sigle quote `'` too....

shu-mutou:
@tyranron PTAL again.

eakraly:
I feel like we are adding inconsistency: we can pass multiple arguments in command line, but those 4 specific are through environment variables.
What about using arguments instead (something like --detect-external-ip) which will signal the entrypoint to do exactly the same?

tyranron:
> To use it from shell script for CD , I also use envs in `--mysqldb-userdb` like `--mysqldb-userdb="host=${MYSQL_HOST} dbname=${MYSQL_DBNAME} user=${MYSQL_USER} password=${MYSQL_PASS} port=${MYSQL_PORT}"`. So current implement is over expanding for me.

@shu-mutou could you specify the whole command you're using, so I can tackle with it and try to think of the solution?

shu-mutou:
@tyranron 

I tried following command.

```
docker run coturn/coturn --mysqldb-userdb='host=192.168.1.100 dbname=coturn user=coturn' -V
```

Then, it returns following message at first.

```
turnserver: unrecognized option '--mysqldb-userdb=host=192.168.1.100'
```

Thanks!

shu-mutou:
@eakraly 

> What about using arguments instead (something like --detect-external-ip) which will signal the entrypoint to do exactly the same?

It sounds good! But it's been 20 years since I used C, and I've forgotten it.... sorry.

eakraly:
> @eakraly
> 
> > What about using arguments instead (something like --detect-external-ip) which will signal the entrypoint to do exactly the same?
> 
> It sounds good! But it's been 20 years since I used C, and I've forgotten it.... sorry.

Oh - it has nothing to do with C actually
What I meant is that entrypoint.sh would accept those as arguments instead of env variables.
You pass them to executable as arguments anyway

We are talking about `getopt` https://stackabuse.com/how-to-parse-command-line-arguments-in-bash/
```
if [ -v DETECT_EXTERNAL_IP ]; then <=== Replace this
  DETECT_EXTERNAL_IP=" --external-ip=$(eval "detect-external-ip")"
fi
```

shu-mutou:
@eakraly, thank you for reply!

I think it is better to explicitly separate the usage of coturn command options and docker image options.
Some people may be using coturn binary directly instead of docker, so I think it will cause confusion.

Your solution requires parsing and editing all of the arguments, including options for coturn, and I'm afraid that implementation will cause further problems.

tyranron (approved):
@shu-mutou sorry for replying late on this.

I've considered your examples and played around quite a while to come up with a solution. For all the tests I did, it expands correctly now. Hope, this will fix your issue too.

Thanks for bringing this up.

tyranron:
@shu-mutou released in [4.6.1-r1](https://github.com/coturn/coturn/releases/tag/docker%2F4.6.1-r1).

shu-mutou:
@tyranron Thank you so much for checking and merging this! :heart:

PR #1114, opened 2022-12-04
Set string bytes to null to prevent random origin

### LINKED ISSUE #1111: Randomly add origin to Redis  because access management page
1. Access  origin  by url:  https://127.0.0.1:8080/os?realm=leedarson
2. Because it is not determined that the add_realm attribute is not null, the randomly allocated memory is not null, resulting in the random addition of origin.

The code is as follows
`  
const uint8_t *add_realm = (const uint8_t*)current_eff_realm();
const uint8_t *add_origin = (const uint8_t*)get_http_header_value(hr, HR_ADD_ORIGIN, "");
const char* msg = "";
uint8_t corigin[STUN_MAX_ORIGIN_SIZE+1];
get_canonic_origin((const char *)add_origin, (char *)corigin, sizeof(corigin)-1);
if(corigin[0]) {
     ....
}
`

if(add_origin  &&  add_origin[0]  &&  corigin[0])

eakraly:
Fix #1111

### DIFF
--- a/src/apps/relay/turn_admin_server.c
+++ b/src/apps/relay/turn_admin_server.c
@@ -3511,7 +3511,7 @@ static void handle_https(ioa_socket_handle s, ioa_network_buffer_handle nbh) {
                 uint8_t o[STUN_MAX_ORIGIN_SIZE + 1];
                 STRCPY(o, origin);
                 dbd->del_origin(o);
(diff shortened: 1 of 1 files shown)


ggarber (approved):
👏 👏 👏

PR #1115, opened 2022-12-04
Add STUN request/response/error prometheus counters

eakraly:
Somewhat relevant to #1075

### DIFF
--- a/src/apps/relay/ns_ioalib_engine_impl.c
+++ b/src/apps/relay/ns_ioalib_engine_impl.c
@@ -3525,21 +3525,23 @@ void stun_report_binding(void *a, STUN_PROMETHEUS_METRIC_TYPE type) {
         if (e && e->verbose) {
           TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "session %018llu: STUN binding %s realm=<%s>, username=<%s>\n",
                         (unsigned long long)ss->id,
-                        (type == 0) ? "request" : (type == 1) ? "response" : "error",
+                        (type == 0)   ? "request"
+                        : (type == 1) ? "response"
+                                      : "error",
                         (char *)ss->realm_options.name, (char *)ss->username);
         }
-        switch(type) {
+        switch (type) {
         case 0:
-            prom_inc_stun_binding_request();
-            break;
-          case 1:
-            prom_inc_stun_binding_response();
-            break;
-          case 2: 
-            prom_inc_stun_binding_error();
-            break;
-          default:
-            break;
+          prom_inc_stun_binding_request();
+          break;
+        case 1:
+          prom_inc_stun_binding_response();
+          break;
+        case 2:
+          prom_inc_stun_binding_error();
+          break;
+        default:
+          break;
         }
       }
     }
--- a/src/apps/relay/ns_ioalib_engine_impl.c
+++ b/src/apps/relay/ns_ioalib_engine_impl.c
@@ -3539,6 +3539,7 @@ void stun_report_binding(void *a, STUN_PROMETHEUS_METRIC_TYPE type) {
             prom_inc_stun_binding_error();
             break;
           default:
+            break;
         }
       }
     }
--- a/src/apps/relay/ns_ioalib_engine_impl.c
+++ b/src/apps/relay/ns_ioalib_engine_impl.c
@@ -3514,6 +3514,63 @@ const char *get_ioa_socket_ssl_method(ioa_socket_handle s) {
   return "no SSL";
 }
 
+void stun_report_binding_request(void *a) {
+  if (a) {
+    ts_ur_super_session *ss = (ts_ur_super_session *)a;
+    if (ss) {
+      turn_turnserver *server = (turn_turnserver *)ss->server;
+      if (server) {
+        ioa_engine_handle e = turn_server_get_engine(server);
+        if (e && e->verbose) {
+          TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "session %018llu: STUN binding request realm=<%s>, username=<%s>\n",
+                        (unsigned long long)ss->id, (char *)ss->realm_options.name, (char *)ss->username);
+        }
+#if !defined(TURN_NO_PROMETHEUS)
+        { prom_inc_stun_binding_request(); }
+#endif
+      }
+    }
+  }
+}
+
+void stun_report_binding_response(void *a) {
+  if (a) {
+    ts_ur_super_session *ss = (ts_ur_super_session *)a;
+    if (ss) {
+      turn_turnserver *server = (turn_turnserver *)ss->server;
+      if (server) {
+        ioa_engine_handle e = turn_server_get_engine(server);
+        if (e && e->verbose) {
+          TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "session %018llu: STUN binding response realm=<%s>, username=<%s>\n",
+                        (unsigned long long)ss->id, (char *)ss->realm_options.name, (char *)ss->username);
+        }
+#if !defined(TURN_NO_PROMETHEUS)
+        { prom_inc_stun_binding_response(); }
+#endif
+      }
+    }
+  }
+}
+
+void stun_report_binding_error(void *a) {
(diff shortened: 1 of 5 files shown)


eakraly (commented):
fix

eakraly:
@paulkram wondered if you could take a look at this change as well? Thank you!

src/apps/relay/prom_server.c
```diff
@@ -41,7 +45,15 @@ void start_prometheus_server(void) {
     nlabels++;
   }
 
-  // Create traffic counter metrics
+  // Create STUN counters
+  stun_binding_request = prom_collector_registry_must_register_metric(
+      prom_counter_new("turn_traffic_rcvp", "Incoming STUN Binding requests", 0, NULL));
+  stun_binding_response = prom_collector_registry_must_register_metric(
+      prom_counter_new("turn_traffic_rcvp", "Outgoing STUN Binding responses", 0, NULL));
```
ggarber, line 52: shouldn't it be _sentp instead of _rcvp if it is for "Outgoing"

src/apps/relay/ns_ioalib_engine_impl.c
```diff
@@ -3514,6 +3514,63 @@ const char *get_ioa_socket_ssl_method(ioa_socket_handle s) {
...
+#if !defined(TURN_NO_PROMETHEUS)
+        { prom_inc_stun_binding_response(); }
+#endif
+      }
+    }
+  }
+}
+
+void stun_report_binding_error(void *a) {
+  if (a) {
```
ggarber, line 3556: can we reduce the code duplication in these 3 new methods (stun_report_xxxx) ?

src/apps/relay/prom_server.c
```diff
@@ -41,7 +45,15 @@ void start_prometheus_server(void) {
     nlabels++;
   }
 
-  // Create traffic counter metrics
+  // Create STUN counters
+  stun_binding_request = prom_collector_registry_must_register_metric(
+      prom_counter_new("turn_traffic_rcvp", "Incoming STUN Binding requests", 0, NULL));
```
ggarber, line 50: the name of the metrics cannot be the same for the 3 metrics, right?

ggarber (commented):
Binding request are supposed to be very lightweight and not be very likely to fail (they don't have auth or require permissions) so I have a small concern that these metrics could be a bad idea in terms of performance implications.  

I think there are use cases with hundreds of thousands stun req/second and the are sync primitives (atomic exchange) inside `prom_counter_add` so I think it is very likely that this would degrade the performance in those cases.

src/apps/relay/prom_server.c
paulkram: I'd actually use the same name but different labels. That should make the dev's live easier when drawing a dashboard and reading the metrics.

I'd suggest:
`const char* label[] = { "type" };
stun_binding = prom_collector_registry_must_register_metric(
      prom_counter_new("turn_traffic_stun", "STUN Binding requests", 1, label));`

usage:
`const char* type[] = { "error" };
prom_counter_add(stun_binding_request, 1, type);`

This should result in way nicer (and less!) metrics on the prometheus DB and thus give performance.

Disclaimer:
Probably 
`stun_binding = prom_collector_registry_must_register_metric(
      prom_counter_new("turn_traffic_stun", "STUN Binding requests", 1, (const char* []) { "type" }));`
  
should also work BUT my C knowledge is quite rusty.

paulkram: Thought a bit about it.

This:
`stun_binding = prom_collector_registry_must_register_metric( prom_counter_new("turn_traffic_stun", "STUN Binding requests", 1, (const char* []) { "type" }));`

Is also broken. Do not use that! It's again an "taking address of temporary array" error when compiling with C++ compiler flags.

src/apps/relay/prom_server.c
eakraly: Names are completely wrong :)

eakraly: Renamed

eakraly:
> Binding request are supposed to be very lightweight and not be very likely to fail (they don't have auth or require permissions) so I have a small concern that these metrics could be a bad idea in terms of performance implications.
> 
> I think there are use cases with hundreds of thousands stun req/second and the are sync primitives (atomic exchange) inside `prom_counter_add` so I think it is very likely that this would degrade the performance in those cases.

Do not disagree. The issue is that today everyone is blind to amount of requests (maybe from network level but only somewhat representative if STUN only mode - with stun/turn traffic impossible to know)

I suggest to have a flag to disable (or enable) stun monitoring as an alternative: if someone wants stun metrics then can enable it.

paulkram:
@eakraly I'm not sure if you've seen my comment. I'm ok with your change and I'd suggest to use labels for the three different STUN metrics instead of three metrics. That would increase performance on promtheus DB a lot.

eakraly:
> @eakraly I'm not sure if you've seen my comment. I'm ok with your change and I'd suggest to use labels for the three different STUN metrics instead of three metrics. That would increase performance on promtheus DB a lot.

Thanks @paulkram - I saw it. My reason not to merge them into one counter (with 3 labels) are following
1. It feels wrong - labels can be used for different realms so we can see if app A makes more requests (or errs) compared to app B. 
2. Performance: I do not know (I admit I did not look into code to compare). Can you please explain how is the performance better with multiple labels vs multiple counters?

paulkram:
@eakraly I'm totally fine, I only wanted to raise awareness that labels might be the better choice.

> It feels wrong - labels can be used for different realms so we can see if app A makes more requests (or errs) compared to app B.

Actually, if you know the realm, you could have used both. realm and type as two separate labels on the same counter. I think this would have benefits in drawing the graphs.

> Performance: I do not know (I admit I did not look into code to compare). Can you please explain how is the performance better with multiple labels vs multiple counters?

I actually refered to the prometheus database. I have no clue about the performance of prom_client_c with or w/o labels. In prometheus DB it has benefits during lookup as labels are usually indexed and thus the requests are faster. Still this is no real solid knowledge.

Again, I'm totally fine with the change :)

ggarber (approved):
Still a bit concerned about performance

eakraly:
@ggarber I was planning to add a flag to make stun metrics an opt-in option - that's why I have not committed it myself
Let me do a follow up change

Also would be really interesting to have some real world numbers for the actual performance... Not something I can provide myself as I run stun/turn combination

dsmeytis:
@eakraly @ggarber I am also concerned about a performance.
I don't have real world numbers so far but I've found this issue https://github.com/digitalocean/prometheus-client-c/issues/59
It means that each call of the `prom_counter_inc()` or `prom_counter_add()` in turn calls expensive `prom_map_get_index_internal()`.
My suggestion is to make these metrics optional or to try another approach:
1. Add STUN binding counters to each `struct _turn_turnserver` and increment them manually without sync primitives (because each `turn_turnserver` structure is used in its own thread only).
2. Use custom collector for STUN binding metrics that simply sums up values among all `relay_server`s and sets prom metric only once when metrics are requested.
I can prepare a PR to show you my idea.

eakraly:
Hi @dsmeytis !

A PR is welcome, of course!

I have to say a few things (just my personal ramblings):
- We are making a lot of hand waving in terms of performance (not backed up by data)
- https://github.com/digitalocean/prometheus-client-c/ library is not a good choice looking forward - it seem to be abandoned (even though I reached out to DigitalOcean through my contacts I had no results). I suggest we look into https://github.com/jupp0r/prometheus-cpp instead (I used it for high performance WebRTC SFU with a TON of metrics and it never had any performance issue).

So longer term I would invest time into replacing `prometheus-client-c`....

dsmeytis:
@eakraly Mehh... my idea also ran into the iceberg of the unmaintained https://github.com/digitalocean/prometheus-client-c/ repo..
Their API for custom collectors lacks essential function. I've created a PR https://github.com/digitalocean/prometheus-client-c/pull/75 but I think it will never be considered.

eakraly:
@dsmeytis I suggest to open a new issue with a topic being "Replace prometheus-client-c with prometheus-cpp"

PR #1118, opened 2022-12-08
Add configuration option for TLS 1.3 ciphersuites

sysvinit:
There are two different API's in OpenSSL for configuring TLS ciphers, one for TLS 1.2 and below, and another for TLS 1.3. coturn only calls the TLS 1.2 API when handling the `--cipher-list` configuration option, which means that it's not possible to use non-default ciphersuites with TLS 1.3 connections.

This PR introduces a new `--ciphersuites` option, which calls the appropriate OpenSSL API to allow TLS 1.3 ciphersuites to be configured.


eakraly:
Thank you @sysvinit !

I do not like the idea of having additional configuration option of TLS 1.3 - there is a lot of duplication with no value

It can be configured by the same string: according to openssl - same cipher list/suits can be passed to both APIs and the values that are unknown will be ignored.

So when having TLS 1.3 available and enabled - also call `SSL_set_ciphersuites` with the same list

See https://www.openssl.org/docs/manmaster/man3/OSSL_default_ciphersuites.html#NOTES

Also better use `OSSL_default_cipher_list()` and `OSSL_default_ciphersuites()` as
```
OSSL_default_cipher_list() and OSSL_default_ciphersuites() replace SSL_DEFAULT_CIPHER_LIST and TLS_DEFAULT_CIPHERSUITES, respectively. The cipher list defines are deprecated as of 3.0.
```

eakraly:
>These functions were introduced in OpenSSL 3.0 and do not exist in OpenSSL 1.1

Sure - that's why I stated `also call SSL_set_ciphersuites with the same list`. My mistake was that I thought it applies to TLS 1.3 and not OpenSSL 3.0

src/apps/relay/mainrelay.c
```diff
@@ -3524,12 +3526,21 @@ static void set_ctx(SSL_CTX **out, const char *protocol, const SSL_METHOD *metho
 
   SSL_CTX_set_default_passwd_cb(ctx, pem_password_func);
 
-  if (!(turn_params.cipher_list[0]))
+  if (!(turn_params.cipher_list[0])) {
     strncpy(turn_params.cipher_list, DEFAULT_CIPHER_LIST, TURN_LONG_STRING_SIZE);
+#if TLSv1_3_SUPPORTED
+    strncat(turn_params.cipher_list, ":", TURN_LONG_STRING_SIZE - 1);
```
eakraly, line 3532: Buffer overflow: max size allowed to be appended at this point is `TURN_LONG_STRING_SIZE - strlen(turn_params.cipher_list)`

src/apps/relay/mainrelay.c
```diff
@@ -3524,12 +3526,21 @@ static void set_ctx(SSL_CTX **out, const char *protocol, const SSL_METHOD *metho
...
+    strncat(turn_params.cipher_list, DEFAULT_CIPHERSUITES, TURN_LONG_STRING_SIZE - 1);
```
eakraly, line 3533: Same as above

src/apps/relay/mainrelay.c
sysvinit: Fixed in [4f20d01](https://github.com/coturn/coturn/pull/1118/commits/4f20d01f7eea038e98d9b29c0004a313ca7159e7)?

src/apps/relay/mainrelay.c
sysvinit: Fixed in [4f20d01](https://github.com/coturn/coturn/pull/1118/commits/4f20d01f7eea038e98d9b29c0004a313ca7159e7)?

eakraly (approved):
Thank you for the contribution!

eakraly:
@sysvinit I took the liberty to change the PR description slightly to reflect the change done (following the update you made).

PR #1124, opened 2022-12-12
Implement non-blocking recvfrom on Windows

pando-emil:
This pull request is a split of PR #1061

As @KangLin pointed out in the original PR those sockets should ideally be permanently non-blocking for performance reasons, but they are NOT at the moment.
Someone with more knowledge about the code in dtls_listener.c should probably have a look if it would be feasible to change the sockets to non-blocking already at creation, similar to what is done in udpserver.c...

### DIFF
--- a/src/apps/relay/dtls_listener.c
+++ b/src/apps/relay/dtls_listener.c
@@ -621,8 +621,9 @@ start_udp_cycle:
 
   ssize_t bsize = 0;
 #if defined(WINDOWS)
-  // TODO: implement it!!!
   int flags = 0;
+  u_long iMode = 1;
+  ioctlsocket(fd, FIONBIO, &iMode);
 #else
   int flags = MSG_DONTWAIT;
 #endif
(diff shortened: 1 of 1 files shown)


KangLin (approved):

src/apps/relay/dtls_listener.c
```diff
@@ -633,6 +634,11 @@ static void udp_server_input_handler(evutil_socket_t fd, short what, void *arg)
   int conn_reset = is_connreset();
   int to_block = would_block();
 
+#if defined(WINDOWS)
```
KangLin, line 637: Why add this code?

KangLin (commented):
Why add this code?

KangLin:
@mom040267 @bradleythughes please review.

src/apps/relay/dtls_listener.c
pando-emil: This code is to restore the socket to blocking mode, to mimic the behavior on non-Windows builds.

KangLin (approved):
@eakraly Please merge

PR #1125, opened 2022-12-12
Explicitly state minimum Windows version

pando-emil:
This pull request is a split of PR #1061

Depending on build system the code will be built against different Windows versions (see comments in old PR).

According to https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt you should define _WIN32_WINNT to the oldest supported platform. And since https://github.com/coturn/coturn/issues/1044 states coturn should support >= Windows 7, I have set it to 0x0601.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,13 @@ IF(MSVC)
     add_definitions(-D_CRT_SECURE_NO_WARNINGS)
 ENDIF(MSVC)
 
+IF(WIN32)
+    # Explicitly state minimum Windows version to compile against
+    # This affects which APIs that are available in for example winsock
+    # 0x0601 == Windows 7
+    add_definitions(-D_WIN32_WINNT=0x0601)
+ENDIF(WIN32)
+
 SET(BUILD_SHARED_LIBS OFF CACHE BOOL "Build shared libs")
 if (BUILD_SHARED_LIBS)
     add_definitions(-DBUILD_SHARED_LIBS)


KangLin:
You maybe understand. The macro is not need by owner defined.

KangLin:
The macro is defined by sdk. 
https://learn.microsoft.com/en-us/windows/win32/WinProg/using-the-windows-headers#setting-winver-or-_win32_winnt
understand the note part

KangLin:
Only you are sure that your current sdk version greater your need sdk version. you are defined it. 
or you are sure that your current sdk version greater 0x7001, then you are define it, support xp etc.
But we can't assume it.
So that we are not define it.

KangLin:
You can make this function by CMAKE.
See: https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html#variable:CMAKE_GENERATOR_TOOLSET
https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_PLATFORM.html#variable:CMAKE_GENERATOR_PLATFORM

KangLin:
Yes. This you said can be done with the cmake configuration parameter

pando-emil:
Ahh, interesting! To be honest I don't fully understand how they should be used. Can you post an example?

KangLin:
ag:
Use vs2015:

cmake -T v140

Use vs2015 support:
cmake -T v140_xp

KangLin:
@pando-emil Please close

pando-emil:
The "cmake -T" solution would only work with visual studio environments, not mingw right? At least it doesn't work in a cross compile environment with mingw.

And if it would work with mingw on Windows, and someone decide cross compile should not be supported, it should still be stated in a cmake-file to achive the points in my comments above.

Neustradamus:
To follow

KangLin:
It is recommended to turn off.

KangLin:
> The "cmake -T" solution would only work with visual studio environments, not mingw right? At least it doesn't work in a cross compile environment with mingw.
> 
> And if it would work with mingw on Windows, and someone decide cross compile should not be supported, it should still be stated in a cmake-file to achive the points in my comments above.

@pando-emil This program does not require it. If you are sure you need to, you can pass this value via the CMake parameter.

ag:
```
cmake -D_WIN32_WINNT=
```

Neustradamus:
@pando-emil: Any news about this PR?

KangLin:
Please close!

pando-fredrik:
> Macros starting with "_" are defined and used internally by the compiler. It is not recommended to use it directly in the program.It is recommended to close.

This is wrong, Microsoft explicitly demonstrates and states you *SHOULD* define them yourself, as demonstrated here, just like @pando-emil stated in the top of this PR.

https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt

"**To modify the macros, in a header file (for example, in targetver.h, which is included by some project templates that target Windows), add the following lines.**

C
```
#define WINVER 0x0A00
#define _WIN32_WINNT 0x0A00
```
The macros in the example are set to target every version of the Windows 10 operating system. The possible values are listed in the Windows header file sdkddkver.h, which defines macros for each major Windows version. To build your application to support a previous Windows platform, include WinSDKVer.h. **Then, set the WINVER and _WIN32_WINNT macros to the oldest supported platform before including sdkddkver.h.** Here are the lines from the Windows 10 SDK version of sdkddkver.h that encode the values for each major version of Windows:"

pando-emil:
@Neustradamus Sorry for the long absense!

I still have no better solution than I proposed with the patch. I simply don't agree with KangLin that there is anything wrong with doing as proposed, but if there is no interest in including the patch, I'll just close the PR.

As I tried to explain above, the patch aims at making the build system more robust to different build environments and future contributions. It is not an issue for me personally, I already figured out what I need to fix outside the build system. :)

CMakeLists.txt
```diff
@@ -69,6 +69,13 @@ IF(MSVC)
     add_definitions(-D_CRT_SECURE_NO_WARNINGS)
 ENDIF(MSVC)
 
+IF(WIN32)
+    # Explicitly state minimum Windows version to compile against
+    # This affects which APIs that are available in for example winsock
+    # 0x0601 == Windows 7
+    add_definitions(-D_WIN32_WINNT=0x0601)
```
jonesmz, line 76: Windows 7 is end of life. Same with 8, 8.1, and server 2012.

This should be either Windows 10, or windows server 2016 to properly reflect supported operating systems.

PR #1136, opened 2023-01-08
Fix brew for MacOS CI build

ggarber:
MacOS CI build started to fail in the PRs because when upgrading python it was not possible to override one existing file (related to 2to3 package).    Apparently this happens because MacOS runners some times have python versions not installed/compatible with brew.

Example of the failure: https://github.com/coturn/coturn/actions/runs/3850951324

The proposed workaround is taken from here: https://github.com/actions/runner-images/issues/2322


eakraly (approved):
Great find

PR #1144, opened 2023-01-18
use santisied psql string

r3g-5z:
Noticed the plaintext password of my postgresql server in my coturn logs, but postgresql errors would return the password sanitised. Simple fix to log the sanitised string.

image

Signed-off-by: r3g_5z <june@girlboss.ceo>


eakraly:
Thanks @r3g-5z !

Issue #1167, opened 2023-03-08
prometheus: cannot use epoll


thib-ack:
Hello,

In my setup, I encountered the issue #997 : when all the 1024 first FD are used, then the /metrics endpoint is broken due to select() usage internally.

so I tried the new version of code which try to use epoll if supported : 
https://github.com/coturn/coturn/blob/79fb65519c25b7d948764bcef73ae92aa1f129b7/src/apps/relay/prom_server.c#L108-L117

However it seems the ```#if MHD_USE_EPOLL_INTERNAL_THREAD``` broke the usage of epoll. On my plateform, with this code, I fallback to ```MHD_USE_SELECT_INTERNALLY; // ubuntu 16.04 ```, but when I remove this check, epoll works greats.

I can see this has been added in #999 as a CI hack

Could you check on your side ? Maybe we can add a log to show which function is used ?

Thanks,
Thibaut.

eakraly:
Hi @thib-ack - I can confirm the code is not perfect.

Could you submit a PR with a fix - something that would work on your system (and not fail the tests)?

eakraly:
Will be released with 4.6.2

PR #1173, opened 2023-03-16
Prometheus: make sure microhttpd starts using epoll if supported

thib-ack:
Hello,

In some cases the prometheus server was started using SELECT even if EPOLL was supported. 
Some flags were changed in microhttpd and now we use MHD_VERSION to make sure to use the right ones in all cases (support old version, for ubuntu 16.04 for ex).

This fixes the issue #1167

I also added a log to make sure we know which version is used, especially to inform the user that the SELECT version might lead to issues on highly used servers.

Thanks,
Thibaut


eakraly:
Thanks @thib-ack !

PR #1174, opened 2023-03-20
Update version to 4.6.2

eakraly:
Update version number, generate authors and changelog files

Release notes (short summary of changes)
```
- Make sure microhttpd starts using epoll if supported
- Add sessioncount to prometheus metrics
- Add STUN request/response/error prometheus counters
- Cleanup logs on turnserver start
- Fix duplicate stdout log output
- Log threadId to logs to aid in multi-threaded debugging
- Optional build info compiled into turnserver binary
- Fix arguments expansion in `docker-entrypoint.sh`
- Santise database connection strings before printing to log
- Support Windows MSVC
- Add configuration option for TLS 1.3 ciphersuites
- Improve openssl3 and FIPS support
- Use single SSL_CTX for TLS and DTLS support
- Update openssl API use to non-deprecated version
- Set string bytes to null to prevent random origin
- Fix memory corruption on socket close
- Fix packet backlog fifo that processed packets in reverse order in some scenarios
- Fix off-by-one when terminating gcm_nonce
- Fixes to Redis memleaks and socketleaks
- Fix malformed response to mobility refresh request
- Fuzzing support
- Ignore raw UDP if no_udp is enabled
- Better detect availability of SCTP protocol
```


ChangeLog
```diff
@@ -1,82 +1,44 @@
...
-	- Add CI tests in older ubuntu version (#981) (Gustavo Garcia)
-	- Replace bzero with memset (#986) (Pavel Punsky)
-	- Fix a warning (#988) (Pavel Punsky)
-	- Improve openssl3 and FIPS support (#955) (Pavel Punsky)
-	- Preserve file timestamps when using install(1) (#983) (Robert Scheck)
-	- Fix renegotiation flag for older version of openssl (#978) (Pavel Punsky)
+- Prometheus: make sure microhttpd starts using epoll if supported (#1173) (Thibaut ACKERMANN <thib-ack@users.noreply.github.com>)
+- Fix typo in mainrelay.c (#1169) (Stefan Sundin <git@stefansundin.com>)
+- Remove unused include that breaks OpenBSD (#1165) (Pavel Punsky <eakraly@users.noreply.github.com>)
+- Update Alpine to 3.16.4 version and Debian "bullseye" to 20230208 snapshot to fix CVE-2023-0286 (and friends) in Docker image (tyranron <tyranron@gmail.com>)
```
tyranron, line 7: @eakraly this (and similar stuff) is not related to the Coturn itself, so is redundant in this changelog file.

ChangeLog
```diff
@@ -1,82 +1,44 @@
...
+- Fix arguments expansion in `docker-entrypoint.sh` (#1110, #902) (Shu Muto <shu.mutow@nec.com>)
+- Delete LICENSE.OpenSSL (Pavel Punsky <eakraly@users.noreply.github.com>)
+- Use santisied psql string (#1144) (r3g_5z <june@girlboss.ceo>)
+- Use the actual redis connection string to connect, not the sanitized one (#1141) (Jasper <jasper@jasperhugo.com>)
+- Update CONTRIBUTING.md (Gustavo Garcia <gustavogb@gmail.com>)
```
tyranron, line 12: @eakraly this also seems to be redundant as is not so useful for end-users.

tyranron (changes requested):
I think changelog entries need careful revisiting to exclude redundant/irreleveant stuff from there.

Also, it's not clear why the previous changelist was wiped out, considering this PR targets `master` branch, so all the `master` changes should be relevant too.

eakraly:
@tyranron pls review again after the fixes. Thanks!

ggarber:
Great job!  Thank you for doing this Pavel.

If @tyranron can confirm the docker part should be ok with this change it looks good to me.

Changelist is long but I think we can create Release Notes manually with a summary of those to make sure people get the most important points.    Also we could start using conventional commits (feat: , chore:....) to make it easer to read if you are ok with it.

tyranron (approved):
@eakraly @ggarber I've removed some Docker-related stuff from the changelog being left, and now it seems fine 👍

PR #1182, opened 2023-04-01
Fix MSVC CI build

PrinceChoco:
Removing windows.h inclusion, not needed with ws2tcpip.h. Causes some struct redefinitions.


eakraly:
Hi @PrinceChoco !

Could you please specify what you see (what errors if any) and how it is resolved?
Is it something that does not show up in CI builds?

PR #1190, opened 2023-04-19
Dockerfile: Install `telnet` tool for easier debugging

stefansundin:
Hello,

I have been setting up coturn and I've had to run `apt update && apt install -y telnet` so many times to telnet in to the cli so I figured I'd try to make it easier by just adding it into the docker image.

Feel free to close if this isn't something you want upstream. Thanks!


tyranron (commented):
@stefansundin yup, this seems to be not a common case of this image usage, rather peculiarities of your particular workflow.

I'd recommend building and using your custom image on top of `coturn/coturn`, containing the commonly required tools by your workflows. This is a common practice for Docker images.

Anyway, thanks for spending your time and effort on this!

PR #1198, opened 2023-05-03
Fix build with libressl 3.6+

rozhuk-im:
Tested on FreeBSD 13/stable


eakraly:
Thank you @rozhuk-im !

PR #1222, opened 2023-06-06
Implement a minimum log level option

PaulWay:
The `-v` or `--verbose` options change a few things, but they have almost no effect
on the `turnserver`.  This implements a `log_min_level` setting and a `--log-min-level`
option to allow the user to set a minimum level below which log messages are not
emitted.

The option takes a level *name* — `debug`, `info`, `warning` or `error`
(case-insensitive) — and is available both as a command-line flag and as a
config-file key:

```
turnserver --log-min-level=warning
```
```ini
# turnserver.conf
log-min-level=warning
```

Scale, most to least verbose: `debug` (the default, i.e. current behaviour) →
`info` → `warning` → `error`. An unrecognised value is reported and ignored
rather than silently reinterpreted, so no argument can leave the server mute.

Documented in `README.turnserver`, the `turnserver(1)` man page, and both example
configs. Covered by a new unit test, `tests/test_log_min_level.c`.

---

### Note: this also drops the `CONTROL` log level

The `TURN_LOG_LEVEL_CONTROL` level is removed from the `TURN_LOG_LEVEL` enum.
Nothing in the tree ever logged at that level — it existed only as an enum value
and two `switch` arms in `ns_turn_utils.c` — so no logging behaviour changes.

Two consequences worth flagging for reviewers:

- **It is a public-header/ABI break.** `ns_turn_utils.h` is installed as a public
  header (`${includedir}/turn`), so `TURN_LOG_LEVEL` is shipped API. Removing the
  middle enumerator renumbers `TURN_LOG_LEVEL_WARNING` (3 → 2) and
  `TURN_LOG_LEVEL_ERROR` (4 → 3). Anything already compiled against
  `libturnclient` that passes `TURN_LOG_LEVEL_ERROR` still passes `4`, which is
  now above every level, so its messages would be dropped. This deserves a
  changelog note on release.

- **This is why `--log-min-level` takes a name and not a number.** With the levels
  renumbered, a numeric argument is a moving target across releases: an earlier
  revision of this PR documented `4 = errors only`, which after the enum change
  sat above every level and silenced the server entirely, fatal errors included.
  Level names cannot drift out of sync with the enum that way.

Relatedly, `log_min_level` is typed as `TURN_LOG_LEVEL` rather than `int`.
The enum has no negative enumerators, so it takes an unsigned underlying type;
as an `int`, a negative minimum was promoted in `level < log_min_level` and
filtered out *every* message instead of none.


eakraly:
Thank you @PaulWay !
I think we should review how the logging level should impact `--verbose` and vice versa - it may create unexpected result when both are used.

jelmd:
I think the patch is ok, however, IMHO there is not that much noise at log level > DEBUG. Therefore I think it is sufficient, to set the default log level to INFO and rise it to DEBUG if `verbose` flag is given. E.g. something like  https://github.com/coturn/coturn/commit/9f986e47ad6734e05ab05a67b8a1712b37fe8ddf ...

jonesmz (approved):

eakraly:
Rebased and made a few small changes (deleted control level that was not used and also not standard)

PR #1231, opened 2023-07-04
Return a 400 response to HTTP requests

davel:
For our deployment, it is useful if coturn returns a valid HTTP response to an HTTP request. To do this on the same port as STUN/TURN and without enabling the admin site, I have extended `read_client_connection()` to return a canned HTTP response, in response to an HTTP request, rather than immediately closing the connection.

### DIFF
--- a/src/server/ns_turn_server.c
+++ b/src/server/ns_turn_server.c
@@ -4549,12 +4549,19 @@ static int read_client_connection(turn_turnserver *server, ts_ur_super_session *
 
           ioa_network_buffer_handle nbh_http = ioa_network_buffer_allocate(ss->client_socket->e);
 
-          char *content = "HTTP Not supported.\n";
+          /* HTTP content */
+          char *content = "HTTP not supported.\n";
+
+          /* Measure length of content */
+          int content_length = strlen(content);
+
+          /* Construct full response */
           char buffer[1024];
           snprintf(buffer, sizeof(buffer),
                    "HTTP/1.1 400 %s Not supported\r\nConnection: close\r\nServer: %s\r\nContent-Type: "
                    "text/plain\r\nContent-Length: %d\r\n\r\n%s",
-                   proto, TURN_SOFTWARE, (int)strlen(content), content);
+                   proto, TURN_SOFTWARE, content_length, content);
+
           ioa_network_buffer_set_size(nbh_http, strlen(buffer));
           memcpy(ioa_network_buffer_data(nbh_http), buffer, strlen(buffer));
           send_data_from_ioa_socket_nbh(ss->client_socket, NULL, nbh_http, TTL_IGNORE, TOS_IGNORE, NULL);
--- a/src/server/ns_turn_server.c
+++ b/src/server/ns_turn_server.c
@@ -4501,7 +4501,8 @@ static int read_client_connection(turn_turnserver *server, ts_ur_super_session *
     if (is_stream_socket(st)) {
       if (is_http((char *)ioa_network_buffer_data(in_buffer->nbh), ioa_network_buffer_get_size(in_buffer->nbh))) {
 
-        const char *proto = "HTTP";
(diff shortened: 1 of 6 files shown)


eakraly:
Thank you @davel !

Since this is a very specific use case we should have a flag to enable this functionality (and have it disabled by default).
This will preserve the existing behavior for everyone today and allow you to have your specific behavior with a flag.

Does it make sense?

davel:
Yes, I can add a flag for this, I shall take a look at the weekend.
 
The change *should* be safe to enable in general. It would only break a workflow which expects an empty response to an HTTP request.

davel:
> Thank you @davel !
> 
> Since this is a very specific use case so it should have a flag to enable this functionality (and have it disabled by default). This will preserve the existing behavior for everyone today and allow you to have your specific behavior with a flag.
> 
> Does it make sense?

This is now done, there is now a `unsupported-http-gives-400` parameter.

src/server/ns_turn_server.c
```diff
@@ -4535,6 +4535,29 @@ static int read_client_connection(turn_turnserver *server, ts_ur_super_session *
...
+          } else {
+            TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "%s: %s returning 400 response", __FUNCTION__, proto);
+          }
+
+          ioa_network_buffer_handle nbh_http = ioa_network_buffer_allocate(ss->client_socket->e);
+
+          char *content = "HTTP Not supported.\n";
+          char buffer[1024];
+          snprintf(buffer, sizeof(buffer),
+                   "HTTP/1.1 400 %s Not supported\r\nConnection: close\r\nServer: %s\r\nContent-Type: "
```
eakraly, line 4555: Let's keep http headers before the content, not after

eakraly (changes requested):
I have reservations regarding the setting name: let's find a more generic name
Maybe `respond-http-unsupported`?

davel:
> I have reservations regarding the setting name: let's find a more generic name Maybe `respond-http-unsupported`?

I have change the option (and variable names) to --respond-http-unsupported.

src/server/ns_turn_server.c
davel: > Let's keep http headers before the content, not after

That would be neater, however we need to know the length of the content before we can generate the header.

src/server/ns_turn_server.c
```diff
@@ -4535,6 +4535,29 @@ static int read_client_connection(turn_turnserver *server, ts_ur_super_session *
...
+                   "text/plain\r\nContent-Length: %d\r\n\r\n%s",
+                   proto, TURN_SOFTWARE, (int)strlen(content), content);
```
ggarber, line 4557: Are you sure you want to disclose the TURN_SOFTWARE here?    I would remove it for any server in production so probably it is better to never include it in this response.

ggarber:
The change looks ok to me.  I would review the issue with disclosing the server name and version for security reasons given that this is in the public interface.

ggarber (approved):

davel:
I added the software version for consistency with the the other http listeners in coturn.  It is not necessary for my application.

davel:
I am happy for the version and server name to be removed from the response. I don't have hardware with me to do so myself for a few days.

src/server/ns_turn_server.c
eakraly: You can generated the body, measure its length, and then construct the response

davel: > You can generated the body, measure its length, and then construct the response

@eakraly  I have added comments and moved the length into a variable to make it clear that it is doing this in that order.

davel:
@eakraly I believe I have resolved all the issues you raised and CI is passing again since merging in current `master`. Is there anything else I need to do to this PR for it to be accepted?

eakraly (approved):

PR #1250, opened 2023-08-07
Fix recursive call in delete alternate server

Cybermilitia:
I tested the code part by using manual and automated tools for more than 2 years.


eakraly:
Thank you for your PR @Cybermilitia !

Could you please expand the description of the PR with what is the issue and how your change fixes it? TY!

Cybermilitia:
- When you try to delete an alternate server in the list by using "das <sever>:<port>" command on the CLI, the session gets stuck and can't response although alternate server is removed. 

- The problem appears to be related to calling del_alt_server function in the same function recursively. 

- I'm trying to eliminate the problem by removing this call.

Issue #1239, opened 2023-07-26
Web UI freeze if no_tls option used


udovichenkoAlexander:
https to web ui freeze in browser if I use **no_tls** option.
Maybe it's ok, because no tls stuff initialized, but no warning or comment in turnconfig mention that. 
I have confused and spent a lot of time to figure it out, because I thought that no_tls option influence in turn/stun only.

eakraly:
Hi @udovichenkoAlexander 
Could you please provide specific example of how you run the server and how you make the quiery and what you see?
That will help with understanding of what happened.

Also: yes - no_tls is global and all incoming connections stop supporting TLS once it is enabled

udovichenkoAlexander:
try add next options to config

```
web-admin
web-admin-port=8080
no-tls
```
next try open web-ui with ```curl -k https://127.0.0.1:8080```
command freeze (no response to tcp SYN packet)

if *no-tls* option commented 
```
web-admin
web-admin-port=8080
#no-tls
```

Now web ui works ```curl -k  https://127.0.0.1:8080```
```
<!DOCTYPE html>
<html>
  <head>
    <title>TURN Server (https admin connection)</title>
 <style> table, th, td { border: 1px solid black; border-collapse: collapse; text-align: left; padding: 5px;} table#msg th { color: red; background-color: white; } </style> </head>
  <body>
    <b>TURN Server</b><br><i>https admin connection</i><br>
<br><br>
<form action="/logon" method="POST">
  <fieldset><legend>Admin user information:</legend>  user name:<br><input required type="text" name="uname" value=""><br>password:<br><input required type="password" name="pwd" value=""><br><br><input type="submit" value="Login"></fieldset>
</form>
</body>
</html>
```

So no-tls (global) and web-ui (https only allowed) is incompatible. At least comment will be helpfull

```
# Enable Web-admin support on https. By default it is Disabled.
# If it is enabled it also enables a http a simple static banner page
# with a small reminder that the admin page is available only on https.
-> # no-tls option require
web-admin

```

eakraly:
@udovichenkoAlexander 
Thank you for details!
Feel free to submit a PR with a suggestion on how it can work better and/or document it

PR #1256, opened 2023-08-22
Add warning and disable web admin if no-tls option used

udovichenkoAlexander:
Fixes https://github.com/coturn/coturn/issues/1239

https to web ui freeze in browser if no_tls option used, because no tls stuff initialized.
This PR add warning about this and comment aboute this in default config


eakraly:
Thank you @udovichenkoAlexander for the PR!

Could you please add a little more explanation to the description of PR?
Copy paste of a few lines from the issue would be great!

PR #1278, opened 2023-09-30
Fix memleak in pgsql_reread_realms

### LINKED ISSUE #1259: Memory leak in function 'pgsql_reread_realms'
The memory leak is reported by a static analyzer tool developed at [CAST](https://www.linkedin.com/company/cast-center)
Specifically, memory is allocated in [here](https://github.com/coturn/coturn/blob/2a695ea8551e782169c495d9239b2a0388763879/src/apps/relay/dbdrivers/dbd_pgsql.c#L779)
and in the next line when it calls function 'ur_string_map_put' to add the allocated pointer into map, the adder function may fail but it not checked

eakraly:
Fixes #1259

If `ur_string_map_put ` fails then the string that was just `strdup`-ed will leak memory 
Now the return value is checked and memory free-ed in case of failure

### DIFF
--- a/src/apps/relay/dbdrivers/dbd_pgsql.c
+++ b/src/apps/relay/dbdrivers/dbd_pgsql.c
@@ -777,7 +777,10 @@ static void pgsql_reread_realms(secrets_list_t *realms_list) {
             if (rval) {
               get_realm(rval);
               ur_string_map_value_type value = strdup(rval);
-              ur_string_map_put(o_to_realm_new, (ur_string_map_key_type)oval, value);
+              int ret = ur_string_map_put(o_to_realm_new, (ur_string_map_key_type)oval, value);
+              if (ret == -1) {
+                free(rval);
+              }
             }
           }
         }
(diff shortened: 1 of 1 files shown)


src/apps/relay/dbdrivers/dbd_pgsql.c
```diff
@@ -777,7 +777,10 @@ static void pgsql_reread_realms(secrets_list_t *realms_list) {
             if (rval) {
               get_realm(rval);
               ur_string_map_value_type value = strdup(rval);
-              ur_string_map_put(o_to_realm_new, (ur_string_map_key_type)oval, value);
+              int ret = ur_string_map_put(o_to_realm_new, (ur_string_map_key_type)oval, value);
+              if (ret == -1) {
+                free(rval);
```
ggarber, line 782: Thx for the fix @eakraly 
aren't we supposed to free `value` instead of `rval`?

ggarber (changes requested):

src/apps/relay/dbdrivers/dbd_pgsql.c
eakraly: Right, why did I write this?

PR #1300, opened 2023-10-23
Windows: add windows service

KangLin:
- Add windows service interface

PR #1301, opened 2023-10-24
Refactor: Refactor the mainrelay.c

KangLin:
Refactor the mainrelay.c for add windows service.
See #1300

- Replace exit() with an elegant way
- Add free resource when the program exit

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -2854,7 +2854,7 @@ static void set_network_engine(void) {
 #endif /* defined(SO_REUSEPORT) */
 }
 
-static void drop_privileges(void) {
+static int drop_privileges(void) {
 #if defined(WINDOWS)
   // TODO: implement it!!!
 #else
@@ -2863,7 +2863,7 @@ static void drop_privileges(void) {
     if (getgid() != procgroupid) {
       if (setgid(procgroupid) != 0) {
         perror("setgid: Unable to change group privileges");
-        exit(-1);
+        return -1;
       } else {
         TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "New GID: %s(%lu)\n", procgroupname, (unsigned long)procgroupid);
       }
@@ -2876,7 +2876,7 @@ static void drop_privileges(void) {
     if (procuserid != getuid()) {
       if (setuid(procuserid) != 0) {
         perror("setuid: Unable to change user privileges");
-        exit(-1);
+        return -1;
       } else {
         TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "New UID: %s(%lu)\n", procusername, (unsigned long)procuserid);
       }
@@ -2885,6 +2885,7 @@ static void drop_privileges(void) {
     }
   }
 #endif
+  return 0;
 }
 
 static void init_domain(void) {
@@ -3930,6 +3931,7 @@ long service_start(int argc, char **argv) {
 }
 
 long service_run() {
+  long lRet = 0;
 
 #if !defined(WINDOWS)
   if (turn_params.pidfile[0]) {
@@ -3982,7 +3984,8 @@ long service_run() {
   event_add(ev, NULL);
 #endif
 
-  drop_privileges();
+  lRet = drop_privileges();
+  if(lRet) return lRet;
 
   start_prometheus_server();
 
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -104,7 +104,7 @@ turn_params_t turn_params = {
     0, /*no_tlsv1*/
     0, /*no_tlsv1_1*/
     0, /*no_tlsv1_2*/
-       /*no_tls*/
+/*no_tls*/
 #if !TLS_SUPPORTED
     1,
 #else
@@ -234,6 +234,9 @@ turn_params_t turn_params = {
     0, /* response_origin_only_with_rfc5780 */
     0  /* respond_http_unsupported */
 };
+
+// Initial the common of turnadmin and turnserver
+int init_common(int argc, char **argv);
 
 //////////////// OpenSSL Init //////////////////////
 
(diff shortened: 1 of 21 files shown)


KangLin:
@SamuelMarks @eakraly @rolandwu777 @ggarber @pando-emil 

Please review.

src/apps/relay/ns_ioalib_engine_impl.c
```diff
@@ -3798,6 +3798,19 @@
...
+  if(!r) return;
+  TURN_MUTEX_LOCK(&r->mutex_sm);
+  int i = 0;
+  for (i = 0; i <= r->sm_chunk; ++i) {
+    free(r->super_memory[i]);
+    free(r->super_memory);
+    free(r->sm_allocated);
+  }
+  free(r);
+  TURN_MUTEX_LOCK(&r->mutex_sm);
```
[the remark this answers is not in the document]
SamuelMarks: @KangLin just move this one line higher or maybe it needs to removed entirely

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
 
-  optind = 0;
+static DH *get_dh1066(void) {
 
-  while (((c = getopt_long(argc, argv, OPTIONS, uo.u.o, NULL)) != -1)) {
-    if (c == 'u') {
-      set_option(c, optarg);
-    }
-  }
+  unsigned char dh1066_p[] = {0x02, 0x0E, 0x26, 0x6F, 0xAA, 0x9F, 0xA8, 0xE5, 0x3F, 0x70, 0x88, 0xF1, 0xA9, 0x29, 0xAE,
```
jonesmz, line 3118: static, const?

also, perhaps use uint8_t instead of "unsigned char" ?

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
-    TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING, "CONFIG: WARNING: --server-relay: NON-STANDARD AND DANGEROUS OPTION.\n");
+static int pem_password_func(char *buf, int size, int rwflag, void *password) {
+  UNUSED_ARG(rwflag);
+
+  strncpy(buf, (char *)(password), size);
+  buf[size - 1] = 0;
+  return (strlen(buf));
+}
+
+#if ALPN_SUPPORTED
```
jonesmz, line 3205: it would be really nice to have all these ALPN functions moved to their own implementation file.

There's not really any reason to have everything all in this one implementation file.

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
+    int set_auto_curve = 0;
 
-    if (!turn_params.relays_number) {
-      TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "%s: You must specify the relay address(es)\n", __FUNCTION__);
-      fprintf(stderr, "\n%s\n", Usage);
-      exit(-1);
+    const char *curve_name = turn_params.ec_curve_name;
+
+    if (!(curve_name[0])) {
+#if !SSL_SESSION_ECDH_AUTO_SUPPORTED
```
jonesmz, line 3324: shouldn't you check that the macro is defined before checking it's value?

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
-  if (turn_params.turn_daemon) {
-#if !defined(TURN_HAS_DAEMON)
-    pid_t pid = fork();
-    if (pid > 0)
-      exit(0);
-    if (pid < 0) {
-      TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "Cannot start daemon process\n");
-      exit(-1);
+    if (!dh) {
+      if (turn_params.dh_key_size == DH_566)
```
jonesmz, line 3381: braces please

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
+        dh = get_dh2066();
     }
-#else
-    if (daemon(1, 0) < 0) {
-      TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "Cannot start daemon process\n");
-      exit(-1);
+
+    if (!dh) {
+      TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "%s: ERROR: cannot allocate DH suite\n", __FUNCTION__);
+      err = 1;
```
jonesmz, line 3391: shouldn't errors be negative?

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
-  struct event *ev = evsignal_new(turn_params.listener.event_base, SIGUSR2, reload_ssl_certs, NULL);
-  event_add(ev, NULL);
+  {
+    int op = 0;
 
-  ev = evsignal_new(turn_params.listener.event_base, SIGTERM, shutdown_handler, NULL);
-  event_add(ev, NULL);
-  ev = evsignal_new(turn_params.listener.event_base, SIGINT, shutdown_handler, NULL);
-  event_add(ev, NULL);
+#if defined(SSL_OP_NO_SSLv2)
```
jonesmz, line 3426: SSLv2 and v3 are even older than TLS 1.0 and 1.1, which are both deprecated.
Is it a good idea to support old, known insecure, algorithms?

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
+    TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING, "\nWARNING: certificate file is not specified, I cannot start TLS/DTLS "
+                                          "services.\nOnly 'plain' UDP/TCP listeners can be started.\n");
+    turn_params.no_tls = 1;
+    turn_params.no_dtls = 1;
   }
 
-keyerr : {
-  if (critical) {
+  if (!(turn_params.no_tls && turn_params.no_dtls) && !turn_params.pkey_file[0]) {
+    TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING, "\nWARNING: private key file is not specified, I cannot start TLS/DTLS "
```
jonesmz, line 3509: in my opinion, if the certificate file isn't specified, but dtls / tls is configured to be turned on, then the process should exit. There's no reason to allow a misconfiguration like this to continue.

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
-static void adjust_key_file_names(void) {
-  if (turn_params.ca_cert_file[0])
-    adjust_key_file_name(turn_params.ca_cert_file, "CA", 1);
-  adjust_key_file_name(turn_params.cert_file, "certificate", 1);
-  adjust_key_file_name(turn_params.pkey_file, "private key", 1);
-  if (turn_params.dh_file[0])
-    adjust_key_file_name(turn_params.dh_file, "DH key", 0);
-}
-static DH *get_dh566(void) {
+  if (!(turn_params.no_tls && turn_params.no_dtls)) {
```
jonesmz, line 3515: this would be easier to read at a glance as

`if( ! turn_params.no_tls || ! turn_params.no_dtls)`

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
-  DH *dh;
+static void openssl_load_certificates(void) {
 
-  if ((dh = DH_new()) == NULL)
-    return (NULL);
+  print_abs_file_name("", "Certificate", turn_params.cert_file);
+  print_abs_file_name("", "Private key", turn_params.pkey_file);
+
+  TURN_MUTEX_LOCK(&turn_params.tls_mutex);
+  if (!turn_params.no_tls) {
```
jonesmz, line 3533: boy it's really awkward to see ` ! turn_params.no_tls` over and over.

We should seriously consider making this say `turn_params.tls_enabled` instead, so that the code doesn't need to negate every place that it checks it.

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
+#else  // OPENSSL_VERSION_NUMBER < 0x10100000L
+    set_ctx(&turn_params.dtls_ctx, "DTLS", DTLS_server_method());
+    if (turn_params.no_tlsv1 || turn_params.no_tlsv1_1) {
+      SSL_CTX_set_min_proto_version(turn_params.dtls_ctx, DTLS1_2_VERSION);
+    }
+    if (turn_params.no_tlsv1_2) {
+      SSL_CTX_set_max_proto_version(turn_params.dtls_ctx, DTLS1_VERSION);
+    }
+#endif // OPENSSL_VERSION_NUMBER < 0x10100000L
+    setup_dtls_callbacks(turn_params.dtls_ctx);
```
jonesmz, line 3595: can this function fail? should a return code be checked?

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
+  char config_file[1025] = DEFAULT_CONFIG_FILE;
 
-  const unsigned char *ptr = in;
-  while (ptr < (in + inlen)) {
-    unsigned char current_len = *ptr;
-    if (ptr + 1 + current_len > in + inlen)
+  // The configuration file and log parameters in the command line must be processed first
+  while (((c = getopt_long(argc, argv, OPTIONS, uo.u.o, NULL)) != -1)) {
+    switch (c) {
+    case 'l':
```
jonesmz, line 3681: what is this case for?

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
-      SSL_CTX_set_ecdh_auto(ctx, 1);
-#endif
-#endif
-      set_auto_curve = 0;
+  if (anon_credentials) {
+    if (turn_params.default_users_db.ram_db.users_number) {
+      TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING,
+                    "CONFIG: you specified user accounts, (-u option)	but you also specified the "
+                    "anonymous user access option (-z or --no-auth option).	User accounts will be ignored.\n");
```
jonesmz, line 3849: Instead of ignoring the user accounts flag, this combination should be considered a hard error, and the process should shutdown.

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
+      turn_params.ct = TURN_CREDENTIALS_NONE;
+      use_lt_credentials = 0;
     }
   }
-#endif
 
-  { // DH algorithms:
+  if (use_cli && cli_password[0] == 0) {
+    TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "CONFIG: Empty cli-password, and so telnet cli interface is disabled! "
```
jonesmz, line 3856: this should be considered a hard error and the process should terminate.

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
+  }
+
+  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "=========== Network =========\n");
+  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Default Net Engine version: %d (%s)\n", (int)turn_params.net_engine_version,
+                turn_params.net_engine_version_txt[(int)turn_params.net_engine_version]);
+  int local_listeners = 0;
+  if (!turn_params.listener.addrs_number) {
+    TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING, "NO EXPLICIT LISTENER ADDRESS(ES) ARE CONFIGURED\n");
+    TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "----------- Discovering listener addresses ---------\n");
+    int maddrs = make_local_listeners_list();
```
jonesmz, line 3873: maddrs is an errorcode right? not "member addresses" ?

src/apps/relay/mainrelay.c
```diff
@@ -2779,1082 +2968,1086 @@ static void print_features(unsigned long mfn) {
...
+    local_listeners = 1;
+    TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "-----------------------------------------------------\n");
+    TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Total: %d 'real' addresses discovered\n", maddrs);
+    TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "-----------------------------------------------------\n");
+  }
+
+  if (!turn_params.relays_number) {
+    if (!local_listeners && turn_params.listener.addrs_number && turn_params.listener.addrs) {
+      size_t la = 0;
+      for (la = 0; la < turn_params.listener.addrs_number; la++) {
```
jonesmz, line 3888: can't the size_t be part of the for loop?

src/apps/relay/netengine.c
```diff
@@ -689,6 +690,9 @@ int send_session_cancellation_to_relay(turnsession_id sid) {
 static int handle_relay_message(relay_server_handle rs, struct message_to_relay *sm) {
   if (rs && sm) {
 
+    if (eve(turn_params.verbose))
```
jonesmz, line 693: braces please

src/apps/relay/netengine.c
```diff
@@ -748,10 +752,10 @@ static int handle_relay_message(relay_server_handle rs, struct message_to_relay
       ioa_socket_handle s = sm->m.sm.s;
 
       if (!s) {
-        TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "%s: mobile socket EMPTY\n", __FUNCTION__);
+        TURN_LOG_CATEGORY("relay", TURN_LOG_LEVEL_ERROR, "%s: mobile socket EMPTY\n", __FUNCTION__);
       } else if (s->read_event || s->bev) {
-        TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "%s: mobile socket wrongly preset: 0x%lx : 0x%lx\n", __FUNCTION__,
-                      (long)s->read_event, (long)s->bev);
+        TURN_LOG_CATEGORY("relay", TURN_LOG_LEVEL_ERROR, "%s: mobile socket wrongly preset: 0x%lx : 0x%lx\n",
+                          __FUNCTION__, (long)s->read_event, (long)s->bev);
```
jonesmz, line 758: __FUNCTION__ seems like something that should be populated into the log function call automatically using a macro.

Can the macro be adjusted to always populate the function and line number?

src/apps/relay/netengine.c
```diff
@@ -785,35 +789,36 @@ static void handle_relay_auth_message(struct relay_server *rs, struct auth_messa
...
   struct relay_server *rs = (struct relay_server *)ptr;
 
-  while ((n = evbuffer_remove(input, &sm, sizeof(struct message_to_relay))) > 0) {
-
-    if (n != sizeof(struct message_to_relay)) {
-      perror("Weird buffer error\n");
-      continue;
-    }
+  n = evbuffer_get_length(input);
+  if (n < sizeof(struct message_to_relay)) {
```
jonesmz, line 798: wouldn't this be better as `sizeof(sm)` to ensure a mistake isn't made if `sm` is changed to a different variable type in the future?

src/apps/relay/netengine.c
```diff
@@ -935,6 +940,8 @@ static ioa_engine_handle create_new_listener_engine(void) {
                         &turn_params.redis_statsdb
 #endif
       );
+  if (NULL == e)
+    return e;
```
jonesmz, line 944: why return `e` if `e` is null?

src/apps/relay/netengine.c
```diff
@@ -935,6 +940,8 @@ static ioa_engine_handle create_new_listener_engine(void) {
                         &turn_params.redis_statsdb
 #endif
       );
+  if (NULL == e)
```
jonesmz, line 943: braces please

src/apps/relay/netengine.c
```diff
@@ -943,73 +950,29 @@ static ioa_engine_handle create_new_listener_engine(void) {
 static void *run_udp_listener_thread(void *arg) {
   static int always_true = 1;
```
jonesmz, line 951: erm.... this is not a great name for a variable...?
especially since it seems this specific variable does not accomplish anything.

src/apps/relay/netengine.c
```diff
@@ -1816,57 +2100,67 @@ void setup_server(void) {
...
-  setup_admin_server();
+    nRet = setup_admin_server();
+    if (nRet)
+      break;
 
-  barrier_wait();
+    barrier_wait();
+    return 0;
+
+  } while (0);
```
jonesmz, line 2159: this is a bizare construction. and not really what do{}while(); is for.

Why not use a goto instead?

eakraly:
very old PR with broad scope
Closing
Pls open a new one if there is something still valuable that was in this PR

PR #1315, opened 2023-11-13
CMake: find prometheus(fix #1304)

### LINKED ISSUE #1304: CMake Build doesn't detect Prometheus
I've noticed that the CMake build doesn't seem to detect Prometheus libraries, although configure can.

It seems like a script to find it would be needed based on the output from the CMake github automation: https://github.com/coturn/coturn/actions/runs/6622872744/job/17989011141#step:5:106  (note that the github automation does install prometheus, so it would be expected to find it)

There's some other libraries listed that also don't have a find script. Sorry for no PR on this, CMake scripting is outside my wheelhouse.

KangLin:
fix #1304

### DIFF
--- a/.github/workflows/mingw.yml
+++ b/.github/workflows/mingw.yml
@@ -44,6 +44,7 @@ jobs:
             mingw-w64-x86_64-sqlite3 ^
             mingw-w64-x86_64-hiredis ^
             mingw-w64-x86_64-postgresql ^
+            mingw-w64-x86_64-libmicrohttpd ^
             git base-devel
 
     - name: make_directory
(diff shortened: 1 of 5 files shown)


.github/workflows/cmake.yaml
```diff
@@ -2,7 +2,6 @@ name: CMake
 
 on:
   push:
-    branches: [ $default-branch ]
```
eakraly, line 5: Why change it?

KangLin: Because need run github action in my clone.

eakraly: Let's remove it then and submit.Thanks!

eakraly (approved):
Assuming the default-branch build change is reverted

PR #1319, opened 2023-11-17
Fix: Return correct error code for `create_relay_connection` in case of `RESERVATION-TOKEN` failure

### LINKED ISSUE #1266: `create_relay_connection` doesn't seem to return correct error code for `RESERVATION-TOKEN` failure
As per RFC 5766, [section 6.2](https://www.rfc-editor.org/rfc/rfc5766#section-6.2) point no. 5,
```
5.  The server checks if the request contains a RESERVATION-TOKEN
       attribute.  If yes, and the request also contains an EVEN-PORT
       attribute, then the server rejects the request with a 400 (Bad
       Request) error.  Otherwise, it checks to see if the token is
       valid (i.e., the token is in range and has not expired and the
       corresponding relayed transport address is still available).  If
       the token is not valid for some reason, the server rejects the
       request with a 508 (Insufficient Capacity) error.
```
https://github.com/coturn/coturn/blob/17e3b81a36237352ec3948f5fa1328825e95cf59/src/server/ns_turn_server.c#L4248

If I am not missing anything (quite new to the codebase), the error code should be 508.

Subhra264:
Fixes #1266 

According to RFC 5766, [section 6.2](https://www.rfc-editor.org/rfc/rfc5766#section-6.2) point no. 5, the turn server needs to reject the request with 508 (Insufficient Capacity) error code when the given RESERVATION-TOKEN is not valid.

### DIFF
--- a/src/server/ns_turn_server.c
+++ b/src/server/ns_turn_server.c
@@ -4241,7 +4241,7 @@ static int create_relay_connection(turn_turnserver *server, ts_ur_super_session 
           ioa_socket_tobeclosed(s)) {
 
         IOA_CLOSE_SOCKET(s);
-        *err_code = 404;
+        *err_code = 508;
         *reason = (const uint8_t *)"Cannot find reserved socket";
         return -1;
       }


KangLin (approved):

eakraly:
Hi @Subhra264 !
Thank you for the PR - this is quite an obscure issue that has been hiding in the code for a long time.

I have a few questions:
- How did you find it?
- How did you test it?
I wonder if we could add some tests around this. If we could come up with a test scenario that we could automate then ot could be a base for many more tests that would cover corner cases.

Subhra264:
> How did you find it?

I was just checking the whole RFC 6156 workflow implemented in coturn as a reference for similar implementation in webrtc.rs. While analyzing the code, the error code for the above case seemed incorrect as per the RFC specification.

> - How did you test it?
I wonder if we could add some tests around this. If we could come up with a test scenario that we could automate then ot could be a base for many more tests that would cover corner cases.

Unfortunately, it was a manual discovery and not test-driven. However, I will try exploring ways to add automated tests for different edge cases. Suggestions would be awesome.

eakraly:
Thank you again @Subhra264 !

Maybe creating a script (like in examples folder) that would launch turnserver, peer (if needed) and use turnutil_uclient to simulate the scenario. It is possible that uclient does not support what you need though.

PR #1325, opened 2023-11-23
Refactor: peer_input_handle

KangLin:


src/server/ns_turn_server.c
```diff
@@ -4681,80 +4681,79 @@ static void peer_input_handler(ioa_socket_handle s, int event_type, ioa_net_data
 
   int ilen =
       min((int)ioa_network_buffer_get_size(in_buffer->nbh), (int)(ioa_network_buffer_get_capacity_udp() - offset));
+  if (ilen < 0)
```
jonesmz, line 4684: braces please

src/server/ns_turn_server.c
```diff
@@ -4681,80 +4681,79 @@ static void peer_input_handler(ioa_socket_handle s, int event_type, ioa_net_data
...
+  }
 
-      if (chnum) {
+  if (chnum) {
 
-        size_t len = (size_t)(ilen);
+    size_t len = (size_t)(ilen);
 
-        nbh = in_buffer->nbh;
+    nbh = in_buffer->nbh;
```
jonesmz, line 4710: can the declaration and definition of `nbh` happen on the same line?

src/server/ns_turn_server.c
```diff
@@ -4681,80 +4681,79 @@ static void peer_input_handler(ioa_socket_handle s, int event_type, ioa_net_data
...
-          stun_attr_add_fingerprint_str(ioa_network_buffer_data(nbh), &len);
-          ioa_network_buffer_set_size(nbh, len);
-        }
-      }
-      if (eve(server->verbose)) {
-        uint16_t *t = (uint16_t *)ioa_network_buffer_data(nbh);
-        TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Send data: 0x%x\n", (int)(nswap16(t[0])));
-      }
+    if (!(*server->no_software_attribute)) {
+      const uint8_t *field = (const uint8_t *)get_version(server);
```
jonesmz, line 4738: minor: consider east-const instead of const-west.

`uint8_t const* const field = ....` is much more clear than `const uint8_t * const field = ...`

ggarber: we use const uint8_t* everywhere in the code so we should keep it that way imo

ggarber:
@KangLin Can you apply the two small comments and we merge it quickly?

Thx a lot for improving this code.

src/server/ns_turn_server.c
KangLin: It is possible after C99, but not before.
CMAKE is must supports C11, but MAKEFILE does not necessarily.

KangLin: The reason why it is not defined on the same line is because this variable is used in multiple code blocks

jonesmz: I don't see any value in bothering with c99

PR #1334, opened 2023-12-01
Added sessionID to some log lines

korayvt:
Some events are missed when logs are filtered by session ID. That's why I added the sessionID to some log lines.


korayvt:
Can you please run the checks again? 
I think the failed test is not related with the code I committed.

PR #1342, opened 2023-12-06
increase log timestamp resolution to milliseconds

jelmd:
If --new-log-timestamp is enabled and the --new-log-timestamp-format ends with %N, the time gets formatted as UTC timestamp whereby %N gets replaced by the microseconds part of the current time.


KangLin:
#1344

jelmd:
Yes, but this requires a lot more work, refactoring. So 'til someone gets the time to do it, this fills the gap.

jonesmz (approved):

eakraly:
addressed in #2031

PR #1359, opened 2024-01-02
Avoid string-literal decay to char*

jonesmz:


jonesmz:
Nevermind this, a misunderstanding. I normally use C++, not C.

PR #1362, opened 2024-01-08
Only set MHD_USE_DUAL_STACK if IPv6 is available

zinid:
If IPv6 is not enabled during runtime, prometheus server fails to start with `EAFNOSUPPORT` because `MHD_USE_DUAL_STACK` is set unconditionally.

This PR fixes it. As a bonus, it also checks if libmicrohttpd is compiled with IPv6 support.

### DIFF
--- a/src/apps/relay/prom_server.c
+++ b/src/apps/relay/prom_server.c
@@ -1,6 +1,9 @@
 #include "prom_server.h"
 #include "mainrelay.h"
 #include "ns_turn_utils.h"
+#include <errno.h>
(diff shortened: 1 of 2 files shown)


zinid:
Any chance making it into upstream? Do you see any problems with the PR? Should I improve something?

eakraly:
> Any chance making it into upstream? Do you see any problems with the PR? Should I improve something?

Hi @zinid ! Sorry about delay. Reviewing it now.

Issue #1367, opened 2024-01-12
There is no "test" folder


jonesmz:
This commit https://github.com/coturn/coturn/commit/40c99db6ba154ef52fdead87f31cc2aa4d218c38

Added a BUILD_TEST option to the cmake build, but there is no "test" folder added to the repo.

PR #1370, opened 2024-01-12
Add github action that runs tests with compiler sanitizers

jonesmz:


.github/workflows/compiler-sanitizers.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  compiler-sanitizers:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        sanitizer: [ 'address,pointer-compare,pointer-subtract', 'thread' ]
+
+    env:
+      CFLAGS:       "-fno-omit-frame-pointer -fstack-protector-all -fsanitize=${{matrix.sanitizer}},bounds,enum -fsanitize-address-use-after-scope -fsanitize-address-use-after-return=always -fsanitize-recover=address -fsanitize-memory-track-origins=2"
```
eakraly, line 17: Looking into the log of the action there is a warning:
```
clang: warning: argument unused during compilation: '-fsanitize-memory-track-origins=2' [-Wunused-command-line-argument]
```
Can it be fixed?

eakraly:
Thank you @jonesmz for the contribution!
Could you please explain how it is to be used? Will the action fail when sanitizers detect an issue?

.github/workflows/compiler-sanitizers.yml
jonesmz: > Looking into the log of the action there is a warning:
> 
> ```
> clang: warning: argument unused during compilation: '-fsanitize-memory-track-origins=2' [-Wunused-command-line-argument]
> ```
> 
> Can it be fixed?

Yes, i can address this. It'll just appear to be a little messier in the github actions GUI.

If you prefer the warning be addressed over the GUI looking perfect, it's easy to do.

configure
```diff
@@ -608,10 +608,11 @@ if [ -z "${CC}" ] ; then
         exit 1
     fi
 else
-    ${CC} ${TMPCPROGC} ${OSCFLAGS} -o ${TMPCPROGB} 2>>/dev/null
+    CC_OUTPUT=$(${CC} ${TMPCPROGC} ${OSCFLAGS} -o ${TMPCPROGB})
```
jonesmz, line 611: This change was to aide in understanding why some combinations of sanitizer flags were not working.

The change to this file can be omitted from the pull request now that I have things working. Let me know if you want this file reverted.

jonesmz:
> Thank you @jonesmz for the contribution! Could you please explain how it is to be used? Will the action fail when sanitizers detect an issue?

Yep, you're exactly right.

If any of the programs run by `examples/run_tests.sh` encounters a sanitizer violation, for example, a buffer overflow or out of bounds read would be detected by the address sanitizer, then the execution of the program will halt with an error message printed to the screen.

Note that right now the PR sets `halt_on_error=0` to allow for execution to continue if an error is encountered, but should still return a failure exit code. I did this with the reasoning that there might be many problems with the current code and as a result we wouldn't want to stop on the first problem.

There don't seem to be that many problems though, so it should be no problem to leave `halt_on_error` as the default (=1). Please let me know if that's your preference.

jonesmz:
BTW; important to this PR, there is no "test" folder, even though the CMake code used to reference one. See this recently resolved issue: https://github.com/coturn/coturn/issues/1367

Did this folder ever exist? Having unit tests that are more sophisticated than the `examples/run_tests.sh` script would be extremely helpful for what I'm using coturn for.

configure
eakraly: Got it. I have no strong preference.

eakraly:
> BTW; important to this PR, there is no "test" folder, even though the CMake code used to reference one. See this recently resolved issue: #1367
> 
> Did this folder ever exist? Having unit tests that are more sophisticated than the `examples/run_tests.sh` script would be extremely helpful for what I'm using coturn for.

I do not think so. I looked back in history and it seems like there was intent to add it as part of cmake work on windows but never materialized.

eakraly:
> > Thank you @jonesmz for the contribution! Could you please explain how it is to be used? Will the action fail when sanitizers detect an issue?
> 
> Yep, you're exactly right.
> 
> If any of the programs run by `examples/run_tests.sh` encounters a sanitizer violation, for example, a buffer overflow or out of bounds read would be detected by the address sanitizer, then the execution of the program will halt with an error message printed to the screen.

I see that tests did not actually run because of missing reference to libmicrohttpd.so.2
```
../bin/turnserver: error while loading shared libraries: libmicrohttpd.so.12: cannot open shared object file: No such file or directory
```

> 
> Note that right now the PR sets `halt_on_error=0` to allow for execution to continue if an error is encountered, but should still return a failure exit code. I did this with the reasoning that there might be many problems with the current code and as a result we wouldn't want to stop on the first problem.
> 
> There don't seem to be that many problems though, so it should be no problem to leave `halt_on_error` as the default (=1). Please let me know if that's your preference.
I agree we should see the errors. Let's make the job actually run test though.

jonesmz:
Which tests are you looking at?

the compiler-sanitizer workflow has no errors relayed to libhttp?

https://github.com/coturn/coturn/actions/runs/7507832176/job/20445528483?pr=1370

eakraly (approved):

PR #1387, opened 2024-01-19
CMake: Declare the variable nearby

KangLin:
CMake: Declare the variable nearby﻿


jonesmz:
in my opinion it would be better to have all of the configuration options in one place.

Personally I like to make an `options.cmake` file that gets included at the beginning of the top level `CMakeLists.txt.

KangLin:
Everyone's habits are different. Just keep it the same style

在 2024-01-20 05:41:59，"Michael Jones" ***@***.***> 写道：


in my opinion it would be better to have all of the configuration options in one place.

Personally I like to make an options.cmake file that gets included at the beginning of the top level `CMakeLists.txt.

ggarber:
While I agree with @jonesmz comment I think given that KangLin is the main developer of this file and we are already defining options in the middle of this cmake file it is ok to apply this change and make it consistent.

PR #1390, opened 2024-01-19
Exclude docker alpine s390x since it times out

jonesmz:
It can be re-enabled when it is fixed


jonesmz:
Hrmm, that didn't work. I'll have to double check my syntax.

tyranron:
Merged in https://github.com/coturn/coturn/pull/1395.

PR #1395, opened 2024-01-20
Fix docker test that fails on alpine/s390x

eakraly:


tyranron (approved):
Something bad on QEMU emulation 😕

Incorporated disabling from #1390

PR #1396, opened 2024-01-22
workflow tidying

redraincatching:
pr split on behalf of @jonesmz 

normalisation of some of the workflow files

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,16 +1,23 @@
-name: C/C++ CI MacOS
+name: MacOS
 
 on:
   push:
-    branches: [ $default-branch ]
   pull_request:
     types: [ opened, reopened, synchronize ]
 
 jobs:
   builds:
-    runs-on: macos-12
+    strategy:
+      fail-fast: false
+      matrix: 
+        os: [ 'macos-11', 'macos-12' ]
+    runs-on: ${{ matrix.os }}
+
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
     - name: Install dependencies
       run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/macos.yml
```diff
@@ -1,16 +1,24 @@
...
-    branches: [ $default-branch ]
   pull_request:
     types: [ opened, reopened, synchronize ]
 
 jobs:
   builds:
-    runs-on: macos-12
+    strategy:
+      fail-fast: false
+        # ubuntu 16.04 EOL is 2026
```
eakraly, line 12: Copy paste?

.github/workflows/macos.yml
```diff
@@ -1,16 +1,24 @@
...
+      matrix: 
+        os: [ 'macos-11', 'macos-12' ]
```
eakraly, line 14: Why adding 11?

redraincatching: github supports MacOSX, although if you would rather not support it it can be removed

jonesmz: IMHO coturn should explicitly declare the platforms supported, and out of that list, any that github has runners for should be used.

If Mac OSX 11 isn't supported, then lets declare that to be the case. Else, take advantage of the equipment offered by github.

eakraly: I think one version of mac is enough
Especially if that is older one

ggarber:
I think it could be cleaned up a bit more but it is better than now so LGTM.

As for MacOS versions I don't have a strong opinion so let's see how it goes with both versions.

Probably any decission is better than keeping the PR open longer in this case.

PR #1394, opened 2024-01-19
Add amazon linux tests, tidy up other workflows

jonesmz:
I build and deploy coturn on AmazonLinux2 (EOL in 2025-06), soon to use AmazonLinux2023, for $DayJob.

It would be nice to have coturn's CI also make sure it continues to build on that platform.

While I'm at it, i also try to do some normalization of the various workflow files.
Adds new targets in a couple of places, neatens up some github workflow syntax, so on.

PR #1400, opened 2024-01-25
improvements to and rename of cmake.yaml

redraincatching:
superseding changes made in pr #1394 on request of @jonesmz

dependent on pr #1399 - uses the composite action for dependency installation i wrote there

cleaned up cmake.yml and renamed in the same manner as the other .yml files

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,36 @@
+name: CMake
+
+on:
+  push:
+  pull_request:
+    types: [ opened, reopened, synchronize ]
+
+env:
+  BUILD_TYPE: Release
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      # Even though 16.04 and 18.04 are EOL, don't bother supporting cmake on them. Too old to be worth it.
+      matrix:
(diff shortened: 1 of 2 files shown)


jonesmz:
Looks like this needs to be rebased

jonesmz:
@redraincatching can you rebase this PR?

redraincatching:
the failing build is for ubuntu 18.04, as it is attempting to perform `actions/checkour@v3` with node16, which is no longer officially supported on github runners. as specified [here](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/), there is a temporary workaround, however this will become unusable presently as support for node16 in runners will be removed soon. 

rather than use that, i would recommend a more thorough solution, as @jonesmz has in the [adjust-ubuntu-runners branch](https://github.com/jonesmz/coturn/tree/adjust-ubuntu-runners).

jonesmz:
Can the change applied here be used? https://github.com/coturn/coturn/pull/1548

redraincatching:
@jonesmz yes, that's the temporary fix, and will work until github removes node16 support

PR #1405, opened 2024-01-25
Address clang-tidy warnings in db files

jonesmz:
The general approach here was:

- Always declare variables as close to where they are defined as possible.
- Check for pre-conditions of functions before doing work (e.g. ensure we can connect to the DB before doing a bunch of string formatting)
- Keep the scope of mutexes as reasonably small as practical.
- Use idiomatic C11, such as for-loops over the thing being iterated, not while() loops over constants, or variables that aren't modified.
- Prefer if(fail){return} function-body after over `if(not fail){ function-body inside if} return;

Clang-tidy returns a clean bill of health, but while going through this file i noticed a lot of things that raise questions.

Lack of checking column counts. Lack of handling the possibility of multiple return values. Questionably handling of strings. Complete lack of checking function inputs for invalid values (e.g. nullptr).

I'm not going to fix those, my organization doesn't USE the DB drivers, so i have little interest in re-working the logic beyond addressing clang-tidy warnings for my own sanity, but i did add TODO comments for someone else to look at in the future.

Additional note: While the changes look very invasive.... they aren't.

I don't think there is a way to get github to ignore whitespace in the filediff, but if someone were to compare the commit locally, they'll see that almost all of the changes are just adjusting indentation.

### DIFF
--- a/src/apps/relay/dbdrivers/dbd_mongo.c
+++ b/src/apps/relay/dbdrivers/dbd_mongo.c
@@ -225,11 +225,8 @@ static int mongo_get_user_key(uint8_t *usname, uint8_t *realm, hmackey_t key) {
           char kval[sizeof(hmackey_t) + sizeof(hmackey_t) + 1];
           memcpy(kval, value, sz);
           kval[sz] = 0;
-          if (convert_string_key_to_binary(kval, key, sz / 2) < 0) {
-            TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "Wrong key: %s, user %s\n", kval, usname);
-          } else {
-            ret = 0;
-          }
+          convert_string_key_to_binary(kval, key, sz / 2);
+          ret = 0;
         }
       }
     }
--- a/src/apps/relay/dbdrivers/dbd_sqlite.c
+++ b/src/apps/relay/dbdrivers/dbd_sqlite.c
@@ -44,6 +44,8 @@
 
 #include <pthread.h>
 
+#include <stdbool.h>
+
 //////////////////////////////////////////////////
 
 static pthread_mutex_t rc_mutex = PTHREAD_MUTEX_INITIALIZER;
@@ -53,43 +55,27 @@ static int read_threads = 0;
 static int write_level = 0;
 #if defined(WINDOWS)
 static pthread_t write_thread = {0};
+static pthread_t const null_thread = {0};
 #else
 static pthread_t write_thread = 0;
+static pthread_t const null_thread = 0;
 #endif
 
-static void sqlite_lock(int write) {
-  pthread_t pths = pthread_self();
+static void sqlite_lock(bool const write) {
+  pthread_t const pths = pthread_self();
 
-  int can_move = 0;
+  bool can_move = false;
   while (!can_move) {
     pthread_mutex_lock(&rc_mutex);
-#if defined(WINDOWS)
-    pthread_t zero = {0};
-#endif
     if (write) {
-      if ((
-#if defined(WINDOWS)
-              pthread_equal(write_thread, zero)
-#else
-              write_thread == 0
-#endif
-              && (read_threads < 1)) ||
-          pthread_equal(write_thread, pths)) {
-        can_move = 1;
+      if ((pthread_equal(write_thread, null_thread) && (read_threads < 1)) || pthread_equal(write_thread, pths)) {
+        can_move = true;
         ++write_level;
         write_thread = pths;
       }
-    } else {
-      if ((
-#if defined(WINDOWS)
-              pthread_equal(write_thread, zero))
-#else
-              !write_thread)
-#endif
-          || pthread_equal(write_thread, pths)) {
-        can_move = 1;
-        ++read_threads;
-      }
+    } else if (pthread_equal(write_thread, null_thread) || pthread_equal(write_thread, pths)) {
+      can_move = true;
+      ++read_threads;
     }
     if (!can_move) {
       pthread_cond_wait(&rc_cond, &rc_mutex);
@@ -98,22 +84,15 @@ static void sqlite_lock(int write) {
   }
 }
 
-static void sqlite_unlock(int write) {
+static void sqlite_unlock(bool const write) {
   pthread_mutex_lock(&rc_mutex);
   if (write) {
-    if (!(--write_level)) {
-#if defined(WINDOWS)
-      pthread_t zero = {0};
-      write_thread = zero;
-#else
-      write_thread = 0;
-#endif
+    if (--write_level == 0) {
+      write_thread = null_thread;
       pthread_cond_broadcast(&rc_cond);
     }
-  } else {
-    if (!(--read_threads)) {
-      pthread_cond_broadcast(&rc_cond);
-    }
+  } else if (--read_threads == 0) {
+    pthread_cond_broadcast(&rc_cond);
   }
   pthread_mutex_unlock(&rc_mutex);
 }
--- a/src/apps/relay/dbdrivers/dbd_sqlite.c
+++ b/src/apps/relay/dbdrivers/dbd_sqlite.c
@@ -121,7 +121,6 @@ static void sqlite_unlock(int write) {
 //////////////////////////////////////////////////
 
 static void sqlite_init_multithreaded(void) {
-
 #if defined(SQLITE_CONFIG_MULTITHREAD)
   if (sqlite3_threadsafe() > 0) {
     int retCode = sqlite3_config(SQLITE_CONFIG_MULTITHREAD);
@@ -136,11 +135,8 @@ static void sqlite_init_multithreaded(void) {
     sqlite3_initialize();
   } else {
     TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "Your SQLite database is not compiled to be threadsafe.\n");
-    return;
   }
 #endif
-
-  return;
 }
 
 static int donot_print_connection_success = 0;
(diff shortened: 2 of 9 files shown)


src/apps/relay/dbdrivers/dbd_mongo.c
```diff
@@ -220,11 +220,8 @@ static int mongo_get_user_key(uint8_t *usname, uint8_t *realm, hmackey_t key) {
           char kval[sizeof(hmackey_t) + sizeof(hmackey_t) + 1];
           memcpy(kval, value, sz);
           kval[sz] = 0;
-          if (convert_string_key_to_binary(kval, key, sz / 2) < 0) {
-            TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "Wrong key: %s, user %s\n", kval, usname);
-          } else {
-            ret = 0;
-          }
+          convert_string_key_to_binary(kval, key, sz / 2);
```
jonesmz, line 223: `convert_string_key_to_binary()` never returned anything but 0.
Changed the function to return void to simplify logic at call-sites.

src/ns_turn_defs.h
```diff
@@ -184,6 +184,7 @@ typedef uint32_t turn_time_t;
 #error WRONG BYTE_ORDER SETTING
 #endif
 
+// NOLINTBEGIN(clang-diagnostic-string-compare)
```
jonesmz, line 187: Suppresses clang-tidy warning about comparing the address of constant string literals is implementation specific behavior.

src/apps/relay/dbdrivers/dbdriver.c
```diff
@@ -45,7 +45,7 @@ static void make_connection_key(void) { (void)pthread_key_create(&connection_key
 pthread_key_t connection_key;
 pthread_once_t connection_key_once = PTHREAD_ONCE_INIT;
 
-int convert_string_key_to_binary(char *keysource, hmackey_t key, size_t sz) {
+void convert_string_key_to_binary(char const *keysource, hmackey_t key, size_t sz) {
```
jonesmz, line 48: `convert_string_key_to_binary()` never returned anything but 0.
Changed the function to return void to simplify logic at call-sites.

Also, changed input parameter to `char const*` to allow the optimizer to assume that the contents of `keysource` won't be modified.

src/apps/relay/dbdrivers/dbd_sqlite.c
```diff
@@ -136,19 +135,17 @@ static void sqlite_init_multithreaded(void) {
     sqlite3_initialize();
   } else {
     TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "Your SQLite database is not compiled to be threadsafe.\n");
-    return;
   }
 #endif
-
-  return;
```
jonesmz, line 143: Eliminate clang-tidy warning about unnecessary return at end of void function.

src/apps/relay/dbdrivers/dbd_sqlite.c
```diff
@@ -136,19 +135,17 @@ static void sqlite_init_multithreaded(void) {
...
 }
 
 static int donot_print_connection_success = 0;
 
 static void fix_user_directory(char *dir0) {
   char *dir = dir0;
-  while (*dir == ' ')
+  while (*dir == ' ') {
```
jonesmz, line 146: all control flow should always have braces

src/apps/relay/dbdrivers/dbd_sqlite.c
```diff
@@ -194,15 +191,12 @@ static void init_sqlite_database(sqlite3 *sqliteconnection) {
       "CREATE TABLE admin_user (name varchar(32), realm varchar(127), password varchar(127), primary key (name))",
       NULL};
 
-  int i = 0;
-  while (statements[i]) {
+  for (int i = 0; statements[i] != NULL; ++i) {
```
jonesmz, line 194: initialize variables as close to where they are defined as possible.

in this case, by using a for loop instead of while loop.

src/apps/relay/dbdrivers/dbd_sqlite.c
```diff
@@ -194,15 +191,12 @@ static void init_sqlite_database(sqlite3 *sqliteconnection) {
...
     sqlite3_stmt *statement = NULL;
-    int rc = 0;
-    if ((rc = sqlite3_prepare(sqliteconnection, statements[i], -1, &statement, 0)) == SQLITE_OK) {
+    if (sqlite3_prepare(sqliteconnection, statements[i], -1, &statement, 0) == SQLITE_OK) {
```
jonesmz, line 196: int rc is written to but never read from.

This pattern repeats throughout the whole file.

src/apps/relay/dbdrivers/dbd_sqlite.c
```diff
@@ -212,752 +206,799 @@ static sqlite3 *get_sqlite_connection(void) {
   persistent_users_db_t *pud = get_persistent_users_db();
 
   sqlite3 *sqliteconnection = (sqlite3 *)pthread_getspecific(connection_key);
-  if (!sqliteconnection) {
+  if (sqliteconnection != NULL) {
```
jonesmz, line 209: Exit early if the connection is already available.

src/apps/relay/dbdrivers/dbd_sqlite.c
```diff
@@ -212,752 +206,799 @@ static sqlite3 *get_sqlite_connection(void) {
...
-  sqlite3 *sqliteconnection = get_sqlite_connection();
-  if (sqliteconnection) {
-    char statement[TURN_LONG_STRING_SIZE];
-    sqlite3_stmt *st = NULL;
-    int rc = 0;
-    snprintf(statement, sizeof(statement) - 1, "select value from turn_secret where realm='%s'", realm);
 
-    sqlite_lock(0);
+  sqlite3 * const sqliteconnection = get_sqlite_connection();
+  if (sqliteconnection == NULL) {
```
jonesmz, line 249: exit early if the connection cant be established.

src/apps/relay/dbdrivers/dbd_sqlite.c
```diff
@@ -212,752 +206,799 @@ static sqlite3 *get_sqlite_connection(void) {
...
 
-        int res = sqlite3_step(st);
-        if (res == SQLITE_ROW) {
+	ret = 0;
+    if (sqlite3_column_count(st) > 0) {
 
-          int type = sqlite3_column_type(st, 0);
-          if (type != SQLITE_NULL)
-            add_to_secrets_list(sl, (const char *)sqlite3_column_text(st, 0));
+      for(int stepResult = sqlite3_step(st);
```
jonesmz, line 264: Use a for loop over the result of sqlite3_step() instead of a while-loop over the non-changing `ctotal` (the number of columns) variable.

While the control flow wasn't necessarily... "wrong" before, this is more idomatic.

jonesmz:
The general approach here was:

1. Always declare variables as close to where they are defined as possible.
2. Check for pre-conditions of functions before doing work (e.g. ensure we can connect to the DB before doing a bunch of string formatting)
3. Keep the scope of mutexes as reasonably small as practical.
4. Use idiomatic C11, such as for-loops over the thing being iterated, not while() loops over constants, or variables that aren't modified.
5. Prefer `if(fail){return} function-body after` over `if(not fail){ function-body inside if} return;

src/apps/relay/dbdrivers/dbd_sqlite.c
```diff
@@ -966,208 +991,211 @@ static void sqlite_auth_ping(void *rch) { UNUSED_ARG(rch); }
...
+    for (size_t i = 0; i < rlsz; ++i) {
+      char *realm = realms_list->secrets[i];
 
-          realm_params_t *rp = get_realm(realm);
+      realm_params_t *rp = get_realm(realm);
 
-          lock_realms();
-          rp->options.perf_options.max_bps = turn_params.max_bps;
-          unlock_realms();
+      lock_realms();
```
jonesmz, line 1145: This function was particularly gnarly.

Locking a mutex for a single variable, when there are 2 others to modify as well?

I don't fully know what this function is actually trying to accomplish, but it's a pretty good one for an expert to review.

eakraly:
@jonesmz pls rebase on latest master to fix formatting issues

eakraly:
I took the liberty to rename the PR to more properly reflect the extend of the change @jonesmz

jonesmz:
> @jonesmz pls rebase on latest master to fix formatting issues

Sure, will do on monday.

jonesmz:
> I took the liberty to rename the PR to more properly reflect the extend of the change @jonesmz

I did make modifications to each of the DB  drivers because the function signature of that one function is being changed, but really the change is 99% for the sqlite driver.

Would it be better if i split the PR into two? One for sqlite, the other for the one function?

eakraly:
> > I took the liberty to rename the PR to more properly reflect the extend of the change @jonesmz
> 
> I did make modifications to each of the DB drivers because the function signature of that one function is being changed, but really the change is 99% for the sqlite driver.
> 
> Would it be better if i split the PR into two? One for sqlite, the other for the one function?

It is fine doing all of it in one PR actually. Makes more sense. Just updated the name.
Thank you!

jonesmz:
Most recent commit does a small amount of re-work for the sqlite_lock() and sqlite_unlock() functions to normalize platform specific code differences.  No logic changes, just lifting ifdefs up to the global level instead of littering the functions with the differences.

eakraly:
@jonesmz could you pls rebase the PR? Thanks!

eakraly (approved):

PR #1406, opened 2024-01-26
updated types to bool in _turn_params_ to reflect C11

redraincatching:
approach was as follows, for the `_turn_params_` struct:
- if a variable of type `int` or `vint` was only being used as a boolean, replace it with bool as defined in `<stdbool.h>`
- replace its declaration with true/false, depending on prior assignment as 0/1

changes were only made when i was certain the variables were not being used as an `int`, so i may have missed some

no changes were made to other sections of the code as int-to-bool assignment is allowed in C, only code within the structs were changed, but that can be changed with a later commit

---

from a documentation perspective, it's not clear as to what purpose or benefit the vint alias has. the definition in `ns_turn_defs.h` simply reads

```c
typedef int vint;
typedef vint *vintp;
```
with no comments, and it seems most (but not all) `vint`s are being used as interim booleans through the code. this may just be from lack of knowledge of the codebase, but it doesn't seem useful in any way, so it would be helpful if someone with more expertise could clarify

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -94,7 +94,7 @@ turn_params_t turn_params = {
     false, /*no_tlsv1*/
     false, /*no_tlsv1_1*/
     false, /*no_tlsv1_2*/
-       /*no_tls*/
+           /*no_tls*/
 #if !TLS_SUPPORTED
     true,
 #else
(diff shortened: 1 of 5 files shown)


KangLin (approved):

eakraly:
Thank you @redraincatching for your contribution! coturn code will greatly benefit from modernization

eakraly:
And the lint issue is because of formatting change that just merged - sorry about the rework you need to do!

redraincatching:
@eakraly that formatting issue should be sorted now - glad to hear the changes are being well received. i hope to do the same to a few more files over the next while, too

jonesmz:
@redraincatching can you rebase this PR?

jonesmz (approved):

eakraly:
Hi @jonesmz 
Could you please rebase and see if "mingw / build (windows-latest, Debug, OFF)" can be fixed? Thanks!

redraincatching:
Hi @eakraly, I'm coming back to try and close some of my open PRs - is there a reason why this hasn't been merged? If so, please let me know and I'll resolve the issues.

PR #1408, opened 2024-01-26
Add the InsertBraces command for clang-format to ensure that all conditionals always have braces

jonesmz:
- Why? Because code where conditionals lack braces is much harder to read, and prone to indentation confusion.
- How? Just added an extra flag to .clang-format and re-ran clang-format on all the files.

I also moved .clang-format up to the top level of the repo so that it can be applied to the fuzz targets as well.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -50,7 +50,7 @@ BreakConstructorInitializersBeforeComma: false
 BreakConstructorInitializers: BeforeColon
 BreakAfterJavaFieldAnnotations: false
 BreakStringLiterals: true
-ColumnLimit:     120 
+ColumnLimit:     120
 CommentPragmas:  '^ IWYU pragma:'
 CompactNamespaces: false
 ConstructorInitializerAllOnOneLineOrOnePerLine: false
@@ -133,5 +133,6 @@ StatementMacros:
 TabWidth:        8
 UseCRLF:         false
 UseTab:          Never
+InsertBraces:    true
 ...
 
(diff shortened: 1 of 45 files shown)


.clang-format
```diff
@@ -50,7 +50,7 @@ BreakConstructorInitializersBeforeComma: false
 BreakConstructorInitializers: BeforeColon
 BreakAfterJavaFieldAnnotations: false
 BreakStringLiterals: true
-ColumnLimit:     120 
+ColumnLimit:     120
```
jonesmz, line 53: Remove extra whitespace at end of line.

.clang-format
```diff
@@ -130,8 +130,10 @@ Standard:        Latest
...
   - Q_UNUSED
   - QT_REQUIRE_VERSION
-TabWidth:        8
-UseCRLF:         false
-UseTab:          Never
+TabWidth:           8
+UseCRLF:            false
+UseTab:             Never
+InsertBraces:       true
+InsertNewlineAtEOF: true
```
jonesmz, line 137: Many linux tools get mad if there isn't a single newline at end of file. This helps those tools not get grumpy.

It only works with clang 16 or newer, so the majority of developers probably won't have it, but that's ok, it'll be ignored by versions of clang-format that are too old.

jonesmz: Ugh, nevermind, it breaks the CI.

jonesmz:
Sorry for the force-pushes.
I didn't anticipate arguing with the older clang-format to be so cumbersome.

I'll work on a separate branch if i have an open PR that needs to be tweaked in the future.

eakraly (approved):
Thank you @jonesmz for the contribution! Could not agree more with the format

Issue #1368, opened 2024-01-12
LeakSanitizer: detected memory leaks in turnutils_rfc5769check


jonesmz:
```
make check
  shell: /usr/bin/bash -e {0}
  env:
    CFLAGS:  -fsanitize=address 
    LDFLAGS:  -fsanitize=address 
    CC: clang
bin/turnutils_rfc5769check

=================================================================
==3433==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 45 byte(s) in 1 object(s) allocated from:
    #0 0x559564eaa56e in malloc (/home/runner/work/coturn/coturn/bin/turnutils_rfc5769check+0xab56e) (BuildId: 68a2dfa9ec2e0295ec3342eb3d3da5c02db882c3)
    #1 0x559564eed0cf in base64_encode /home/runner/work/coturn/coturn/src/apps/common/apputils.c:1288:32
    #2 0x559564ee71dc in check_oauth /home/runner/work/coturn/coturn/src/apps/rfc5769/rfc5769check.c:78:7
    #3 0x559564ee6bfb in main /home/runner/work/coturn/coturn/src/apps/rfc5769/rfc5769check.c:560:9
    #4 0x7f1004029d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f) (BuildId: 203de0ae33b53fee1578b117cb4123e85d0534f0)

SUMMARY: AddressSanitizer: 45 byte(s) leaked in 1 allocation(s).
make: *** [Makefile:100: check] Error 1
```

ashamedbit:
Hi,
I am a new contributor looking forward to contribute to the coturn project :).
I was able to reproduce the bug on running `make check` and was able to generate a patch for it. The leak happens within function *check_oauth*. Basically the variable *base64encoded_ltp* is allocated within *base64_encode* however it is not freed  within the coming loop. I have a model patch and I was wondering if I can merge the same. The diff from commit da332ed9e718129d30efc467402d750e3c0fa9a3 is attached as a file below. I was able to verify that after the patch, the leak no longer occurs through leaksanitizer (and there is no double free for that matter)

 Let me know if this is a good patch! I am eager to become a regular contributor to this repository :)
 
 [coturn_leaksanitizer.txt](https://github.com/coturn/coturn/files/14075144/coturn_leaksanitizer.txt)

Neustradamus:
@ashamedbit: Thanks for your comment and your patch!

Can you create directly a PR?
[...]

PR #1410, opened 2024-01-28
Fix memory leak in rfc5769check.c

ashamedbit:
I came across the open issue #1368 which is a leaksanitizer report about a leak that happens on make check. I was able to reproduce the bug on running `make check` and was able to generate a patch for it. The leak happens within function `check_oauth`. 

Issue: The variable `base64encoded_ltp` is allocated within `base64_encode` however it is not de-allocated within the coming loop.

I have verified that after the patch, the leak no longer occurs through `leaksanitizer` (and there is no double free for that matter)

### DIFF
--- a/src/apps/rfc5769/rfc5769check.c
+++ b/src/apps/rfc5769/rfc5769check.c
@@ -189,13 +189,16 @@ static int check_oauth(void) {
   }
 
   if (base64encoded_ltp)
+  {
     free(base64encoded_ltp);
-
+  }
   return 0;
 
   ERROR:
   if (base64encoded_ltp)
-      free(base64encoded_ltp);
+  {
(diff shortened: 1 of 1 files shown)


src/apps/rfc5769/rfc5769check.c
```diff
@@ -157,38 +157,46 @@ static int check_oauth(void) {
...
     }
   }
 
+  if (base64encoded_ltp)
+    free(base64encoded_ltp);
+
   return 0;
+
+  ERROR:
+  if (base64encoded_ltp)
```
jonesmz, line 197: you'll need braces around these if-checks to pass the linter.

ashamedbit: Done!

Issue #1366, opened 2024-01-12
warning: Potential leak of memory pointed to by 'args' [clang-analyzer-unix.Malloc]


jonesmz:
```

/home/runner/work/coturn/coturn/src/apps/relay/netengine.c:350:1: warning: Potential leak of memory pointed to by 'args' [clang-analyzer-unix.Malloc]
}
^
/home/runner/work/coturn/coturn/src/apps/relay/netengine.c:1807:7: note: Assuming 'authserver_number' is >= MIN_AUTHSERVER_NUMBER
  if (authserver_number < MIN_AUTHSERVER_NUMBER)
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/coturn/coturn/src/apps/relay/netengine.c:1807:3: note: Taking false branch
  if (authserver_number < MIN_AUTHSERVER_NUMBER)
  ^
/home/runner/work/coturn/coturn/src/apps/relay/netengine.c:1819:3: note: Calling 'setup_listener'
  setup_listener();
  ^~~~~~~~~~~~~~~~
/home/runner/work/coturn/coturn/src/apps/relay/netengine.c:977:7: note: Assuming field 'ioa_eng' is non-null
  if (!turn_params.listener.ioa_eng)
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/coturn/coturn/src/apps/relay/netengine.c:977:3: note: Taking false branch
  if (!turn_params.listener.ioa_eng)
  ^
/home/runner/work/coturn/coturn/src/apps/relay/netengine.c:980:3: note: Calling 'set_ssl_ctx'
  set_ssl_ctx(turn_params.listener.ioa_eng, &turn_params);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/coturn/coturn/src/apps/relay/netengine.c:335:64: note: Memory is allocated
  update_ssl_ctx_cb_args_t *args = (update_ssl_ctx_cb_args_t *)malloc(sizeof(update_ssl_ctx_cb_args_t));
                                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/coturn/coturn/src/apps/relay/netengine.c:343:7: note: Assuming 'base' is equal to NULL
  if (base != NULL) {
[...]

PR #1411, opened 2024-01-28
Fix memory leak in netengine.c

ashamedbit:
This is in response to issue #1366
The clang static analyzer basically claims that there is a memory leak happening in `set_ssl_ctx` for the variable `args`. The leak is  triggered when the base event `base` is NULL and the condition within `set_ssl_ctx` is not triggered. Therefore as a patch I am adding an else condition to free it. (It cannot be freed after the event is created by `event_new` because `args` can be invoked as argument for callback function later on)

Please let me know if this patch is helpful :)

### DIFF
--- a/src/apps/relay/netengine.c
+++ b/src/apps/relay/netengine.c
@@ -352,7 +352,7 @@ void set_ssl_ctx(ioa_engine_handle e, turn_params_t *params) {
   }
   else
   {
-    if (args) free(args);
+    if (args) { free(args); }
   }
 }
 
--- a/src/apps/relay/netengine.c
+++ b/src/apps/relay/netengine.c
@@ -350,6 +350,10 @@ void set_ssl_ctx(ioa_engine_handle e, turn_params_t *params) {
     params->tls_ctx_update_ev = ev;
     TURN_MUTEX_UNLOCK(&turn_params.tls_mutex);
   }
+  else
+  {
+    if (args) free(args);
+  }
 }
 
 //////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/apps/relay/netengine.c
```diff
@@ -350,6 +350,10 @@ void set_ssl_ctx(ioa_engine_handle e, turn_params_t *params) {
     params->tls_ctx_update_ev = ev;
     TURN_MUTEX_UNLOCK(&turn_params.tls_mutex);
   }
+  else
+  {
+    if (args) free(args);
```
jonesmz, line 355: you'll need to add braces to this if statement to pass the linter.

src/apps/relay/netengine.c
```diff
@@ -350,6 +350,10 @@ void set_ssl_ctx(ioa_engine_handle e, turn_params_t *params) {
     params->tls_ctx_update_ev = ev;
     TURN_MUTEX_UNLOCK(&turn_params.tls_mutex);
   }
+  else
+  {
```
jonesmz, line 354: args is a parameter to the function,

are you certain that all users of the function are expecting the parameter to be freed by this function?

ashamedbit: I believe that the new event added has callback update_ssl_ctx with EV_PERSIST. Therefore,  it doesn't stop being pending when its callback runs. The event can be triggered more than once because it is on persist mode. We can only free this args if the event is not created.

src/apps/relay/netengine.c
ashamedbit: Done!

src/apps/relay/netengine.c
ashamedbit: @jonesmz Thank you for reviewing my PRs! I was wondering if this one still looks good. Also I am unaware of the process to merge the PR into the repository after it has been accepted. Thanks for all the help!

jonesmz: Hi, you're welcome.

I am not able to merge PRs.

Basically you just need to wait for one of the persons with the ability to merge PRs to decide to merge it.

They might ask for additional modifications, or they might merge it as it is.

Issue #1365, opened 2024-01-12
warning: Potential leak of memory pointed to by 'kv' [clang-analyzer-unix.Malloc]


jonesmz:
```
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:172:13: warning: Potential leak of memory pointed to by 'kv' [clang-analyzer-unix.Malloc]
            free(ret);
            ^
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:208:7: note: Assuming 'request' is non-null
  if (request) {
      ^~~~~~~
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:208:3: note: Taking true branch
  if (request) {
  ^
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:212:9: note: Assuming the condition is true
    if (strstr(request, "GET ") == request) {
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:212:5: note: Taking true branch
    if (strstr(request, "GET ") == request) {
    ^
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:214:13: note: Calling 'parse_http_request_1'
      ret = parse_http_request_1(ret, request + 4, 0);
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:153:7: note: 'ret' is non-null
  if (ret && request) {
      ^~~
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:153:7: note: Left side of '&&' is true
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:153:14: note: 'request' is non-null
  if (ret && request) {
             ^~~~~~~
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:153:3: note: Taking true branch
  if (ret && request) {
  ^
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:156:9: note: Assuming 's' is non-null
    if (!s) {
        ^~
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:156:5: note: Taking false branch
    if (!s) {
    ^
/home/runner/work/coturn/coturn/src/apps/relay/http_server.c:163:11: note: Assuming 'uri' is non-null
      if (!uri) {
          ^~~~
[...]

PR #1412, opened 2024-01-28
Fix memory leak on http_server.c

ashamedbit:
This is in response to issue #1365.
The clang static analyzer basically claims that there is a memory leak happening in `parse_http_request_1` for the variable `kv`. The leak is triggered when evhttp_parse_query_str fails and is unable to obtain key value pairs for a given URI. In this case ret is freed, however kv is still not freed and thereafter not used. Therefore as a patch I am freeing kv right after ret is freed.

Please let me know if this patch is helpful :)

### DIFF
--- a/src/apps/relay/http_server.c
+++ b/src/apps/relay/http_server.c
@@ -172,7 +172,7 @@ static struct http_request *parse_http_request_1(struct http_request *ret, char 
           if (evhttp_parse_query_str(query, kv) < 0) {
             free(ret);
             ret = NULL;
-            if (kv) free(kv); //kv no longer assigned on this path
+            if (kv) { free(kv); } //kv no longer assigned on this path
           } else {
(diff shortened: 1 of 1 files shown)


src/apps/relay/http_server.c
```diff
@@ -172,6 +172,7 @@ static struct http_request *parse_http_request_1(struct http_request *ret, char
           if (evhttp_parse_query_str(query, kv) < 0) {
             free(ret);
             ret = NULL;
+            if (kv) free(kv); //kv no longer assigned on this path
```
jonesmz, line 175: you'll need braces around this if-check to pass the linter.

jonesmz (approved):

src/apps/relay/http_server.c
ashamedbit: Done!

eakraly (approved):

PR #1413, opened 2024-01-29
refactored uclient.* to use bool to reflect C11

redraincatching:
similar in nature to my changes in #1406 

approach was as follows:

- replace its declaration with true/false, depending on prior assignment as 0/1

changes were only made when i was certain the variables were not being used as an `int`, so i may have missed some

no changes were made to other sections of the code as int-to-bool assignment is allowed in C, only code within the structs were changed, but that can be changed with a later commit


redraincatching:
i went through the entirety of both files as opposed to keeping the changes contained within a struct as i did previously, but looking at the number of changes should prove that it was still small changes

PR #1420, opened 2024-01-30
Use bool over int for the turnutils_uclient program

jonesmz:
Converts all of the variables in the uclient program that should be bool but weren't.

A few other minor adjustments made at the behest of clang-tidy, but this change does not address all of clang-tidy's complaints.


src/apps/uclient/CMakeLists.txt
```diff
@@ -1,3 +1,37 @@
+##
```
eakraly, line 1: Why are we randomly adding license here?

src/apps/uclient/session.h
```diff
@@ -1,4 +1,6 @@
 /*
+ * SPDX-License-Identifier: BSD-3-Clause
```
eakraly, line 2: I do not understand what this is - let's keep it out of the PR

src/apps/uclient/session.h
```diff
@@ -28,39 +30,36 @@
  * SUCH DAMAGE.
  */
 
-#ifndef __SESSION__
-#define __SESSION__
+#ifndef COTURN_UCLIENT_SESSION_H_3F24DB4E_3426_47FF_BC7F_4F3E1FB3094B
```
eakraly, line 33: Is this necessary in this PR?
Does it conflict with another define?

src/apps/uclient/startuclient.h
```diff
@@ -1,4 +1,6 @@
 /*
+ * SPDX-License-Identifier: BSD-3-Clause
```
eakraly, line 2: as other files

src/apps/uclient/startuclient.h
```diff
@@ -28,10 +30,9 @@
  * SUCH DAMAGE.
  */
 
-#ifndef __STARTCLIENT_TURN__
-#define __STARTCLIENT_TURN__
+#ifndef COTURN_UCLIENT_STARTUCLIENT_H_A1214C56_F7D3_49F2_BA92_AFF74628E01E
```
eakraly, line 33: Let's keep it out of this PR

src/apps/uclient/uclient.c
```diff
@@ -1,4 +1,6 @@
 /*
+ * SPDX-License-Identifier: BSD-3-Clause
```
eakraly, line 2: remove

src/apps/uclient/uclient.h
```diff
@@ -1,4 +1,6 @@
 /*
+ * SPDX-License-Identifier: BSD-3-Clause
```
eakraly, line 2: remove

src/apps/uclient/uclient.h
```diff
@@ -28,15 +30,12 @@
  * SUCH DAMAGE.
  */
 
-#ifndef __UCLIENT_ECHO__
-#define __UCLIENT_ECHO__
+#ifndef COTURN_UCLIENT_UCLIENT_H_BE2A41BA_D261_4A88_9A40_3256EE1BB8DF
```
eakraly, line 33: same

eakraly (changes requested):

eakraly:
Thank you @jonesmz - I left a few comments

src/apps/uclient/CMakeLists.txt
jonesmz: Github terms of use say that if not otherwise indicated, the license for contributions is the default license of the repository.

My reading of https://github.com/coturn/coturn/blob/master/LICENSE is that it's the 3-clause BSD license.

By explicitly listing the license in each file, it helps casual contributors to understand what the license of the project is while they are working.

Going further than that, the SPDX-License-Identifier tag near the top of files is needed for various automated software-bill-of-materials tools to properly understand the provenance of the code.

src/apps/uclient/session.h
jonesmz: This is a software bill of materials tag that helps automated tools understanding the provenance of the code.

This is really important to have.

src/apps/uclient/session.h
jonesmz: Yes, this conflicts with another define in another 3rd party project that my own code uses.

Not namespacing the header guards of a header file with the name of the project is basically guaranteed to have a conflict with something somewhere.

The UUID is also really important, as it provides resilience against having multiple same named files with different contents, which is something that happens a lot with the version control system that my work uses (Perforce), due to the specific way perforce does branching and tagging.

src/apps/uclient/session.h
eakraly: Why should coturn repo add a random software's tag into its code?

src/apps/uclient/session.h
eakraly: How can it conflict with a define in another 3rd party project if this is part of a code of a standalone binary tool of coturn?

src/apps/uclient/session.h
eakraly: And in any case, let's not mix different changes in one PR
In this PR you change ints to bools

jonesmz: > Why should coturn repo add a random software's tag into its code?

SPDX is a standard that's widely used in the open source software ecosystem, it's not a single specific piece of software by some specific vendor. It's an industry standard.

E.g. the Linux kernel uses it, among uncountable many others.

https://www.kernel.org/doc/html/v4.18/process/license-rules.html

jonesmz: > And in any case, let's not mix different changes in one PR In this PR you change ints to bools

Sure, I can split the change.

src/apps/uclient/session.h
jonesmz: `coturn` is not only used in binary form?

I use the code from coturn in source form to compile it into other things.

src/apps/uclient/session.h
jonesmz: split

src/apps/uclient/session.h
jonesmz: I removed the header guard change. It can be applied in a standalone PR some other time.

PR #1421, opened 2024-02-01
changed variables in stunclient.c to bool (C11)

redraincatching:
# changed variables that appeared in `stunclient.c` and their uses to `bool` to follow C11 idioms
## approach was as follows:
- replace its declaration and assignment with true/false, depending on prior assignment as 0/1

changes were only made when i was certain the variables were not being used as an int, so i may have missed some

---

## variables changed in `stunclient.c`
- `rfc5780`
- `change_ip`
- `change_port`

their usages were changed only where they appeared in the apps directory, and then everywhere that generated a warning after make

- `stunclient.c` itself
- `natdiscovery.c`

### DIFF
--- a/src/apps/natdiscovery/natdiscovery.c
+++ b/src/apps/natdiscovery/natdiscovery.c
@@ -28,6 +28,7 @@
  * SUCH DAMAGE.
  */
 
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


jonesmz (approved):

eakraly:
Hi @redraincatching ! Thank you for the contribution!
Did you test it in anyway? The change looks straightforward but we better validate it.

redraincatching:
@eakraly everything has been tested, and is now also `clang-format` compliant

PR #1427, opened 2024-02-09
changed variables in ns_turn_session.h and to bool

redraincatching:
# changed variables that appeared in `ns_turn_session.h` and and their uses to `bool` to follow C11 idioms
## approach was as follows:
- replace its declaration and assignment with true/false, depending on prior assignment as 0/1
- then check any error that appear after build and update them to reflect changes

changes were only made when i was certain the variables were not being used as an int, so i may have missed some

### DIFF
--- a/src/client/ns_turn_msg.c
+++ b/src/client/ns_turn_msg.c
@@ -477,8 +477,8 @@ int old_stun_is_command_message_str(const uint8_t *buf, size_t blen, uint32_t *c
   return 0;
 }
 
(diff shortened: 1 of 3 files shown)


eakraly:
Hi @redraincatching 
I suggest you split PRs - make one change at a time (one variable or one method). It would be much easier to review.

In this PR, you make changes to function definitions but do not change calls to the functions
For example, you change `stun_is_command_message_full_check_str` to accept bool but did not change calls to it. So what you have is silent casting and this may introduce some very subtle errors.

ggarber:
@redraincatching Can you take a look at the comment?

PR #1428, opened 2024-02-09
added missing function prototype of turn_random_number()

redraincatching:
trivial - cmake generated a warning that `ns_turn_msg.c` used a function that had no prior prototype - most other files used the `turn_random()` function but this uses `turn_random_number()` which has no prototype, so i've added it to the header file


eakraly:
Thanks!
We need to review usage and implemetation of `turn_random_number` and `turn_random` - briefly looking at them it looks like one has better implementation but the other one is used throughout the code....  Filed as #1430

PR #1436, opened 2024-02-21
changed int to uint16_t for port variables 

WHYHD:
### Approach:
- Search for variables named port or similar (i.e. _port)
- Swap **int** with **uint16_t** 
- If this effected a method signature search for calls and add **uint16_t** type casting  

This is same scale example to gain feedback on changing ports to unint16_t. I would like to make these changes everywhere so any feedback is welcome.

### DIFF
--- a/docker/coturn/CHANGELOG.md
+++ b/docker/coturn/CHANGELOG.md
@@ -4,6 +4,16 @@ Coturn TURN server Docker image changelog
 
 
 
+## [4.6.2-r11] · 2024-??-?? (unreleased)
+[4.6.2-r11]: /../../tree/docker/4.6.2-r11
+
+### Security updated
+
+- [Debian Linux] "bookworm" 20240701 (12.6): <https://github.com/docker-library/official-images/commit/6ad90d9eb2949d40e88fce34944ccd97752674c8>
+
+
+
+
 ## [4.6.2-r10] · 2024-06-21
 [4.6.2-r10]: /../../tree/docker/4.6.2-r10
 


src/apps/relay/dbdrivers/dbd_redis.c
```diff
@@ -268,7 +268,7 @@ redis_context_handle get_redis_async_connection(struct event_base *base, redis_s
         }
       }
 
-      ret = redisLibeventAttach(base, co->host, co->port, co->password, atoi(co->dbname));
+      ret = redisLibeventAttach(base, co->host, (uint16_t)co->port, co->password, atoi(co->dbname));
```
eakraly, line 271: Pls review if you can change the type to uint16_t in the struct rather than adding cast

eakraly (changes requested):
Left one comment to validate

eakraly:
Thank you @WHYHD for the contribution
The only comment would be to try and get the types right without casting - will be hard to get rid of them later

src/apps/relay/dbdrivers/dbd_redis.c
```diff
@@ -134,9 +134,9 @@ static Ryconninfo *RyconninfoParse(const char *userdb, char **errmsg) {
       } else if (!strcmp(s, "secret")) {
         co->password = strdup(seq + 1);
       } else if (!strcmp(s, "port")) {
-        co->port = (unsigned int)atoi(seq + 1);
+        co->port = atoi(seq + 1);
```
jonesmz, line 137: don't you need a cast to uint16_t to avoid the compiler warning about precision loss?

src/apps/relay/dbdrivers/dbd_redis.c
```diff
@@ -210,7 +210,7 @@ redis_context_handle get_redis_async_connection(struct event_base *base, redis_s
         }
 
         if (co->port) {
-          port = (int)(co->port);
+          port = (co->port);
```
jonesmz, line 213: parenthesis aren't needed here anymore.

PR #1445, opened 2024-03-05
udp-notifier mechanism

alperentokgoz:
Hi All,
It allows retrieving information about failed turn command request to generate some statistics.

This PR provides a general notification interface for the Turn server. It generates informational messages and logs for the Turn server.

The UDP interface is implemented.

The notification interface is currently used for "executed turn command".

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -225,7 +225,9 @@ turn_params_t turn_params = {
     0,                                      /* no_auth_pings */
     0,                                      /* no_dynamic_ip_list */
     0,                                      /* no_dynamic_realms */
-
+    ///// UDP NOTIFIER /////
+    0,
+    "",
     0, /* log_binding */
     0, /* no_stun_backward_compatibility */
     0, /* response_origin_only_with_rfc5780 */
@@ -1217,9 +1219,7 @@ static char Usage[] =
     "						name will be constructed as-is, without PID and date appendage.\n"
     "						This option can be used, for example, together with the logrotate "
     "tool.\n"
-    " --new-log-timestamp				Enable full ISO-8601 timestamp in all logs.\n"
-    " --new-log-timestamp-format    	<format>	Set timestamp format (in strftime(1) format). Depends on "
-    "--new-log-timestamp to be enabled.\n"
+    " --new-log-timestamp				Enable timestamp in all logs.\n"
     " --log-binding					Log STUN binding request. It is now disabled by default to "
     "avoid DoS attacks.\n"
     " --stale-nonce[=<value>]			Use extra security with nonce value having limited lifetime (default "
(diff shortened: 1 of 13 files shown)


src/apps/relay/mainrelay.c
```diff
@@ -2328,6 +2329,10 @@ static void set_option(int c, char *value) {
       turn_params.rest_api_separator = *value;
     }
     break;
+  case UDP_NOTIFIER_OPT:
+    turn_params.udp_notifier = 1;
```
jonesmz, line 2333: please use the C99 "stdbool" header, and the true/false keywords (sometimes implemented as macros by C99) over '1' and '0' and 'int'.

src/apps/relay/mainrelay.h
```diff
@@ -326,6 +326,10 @@ typedef struct _turn_params_ {
   int no_auth_pings;
   int no_dynamic_ip_list;
   int no_dynamic_realms;
+  
+  //// UDP Notifier ////
+  int udp_notifier;
+  char udp_notifier_params[1025];
```
jonesmz, line 332: 1025 is a magic number.

Additionally, does this need to be a flat buffer of characters? can't it be parsed into a struct that has meaningful variable names for whatever the fields are?

jonesmz:
I suppose that mechanically the change is reasonable for what it's trying to accomplish, but i have some concerns about approach and justification

The first observation that I want to make is that this change is hooking only the `handle_turn_command` function to get information like:

1. sessionid
2. origin
3. realm
4. username
5. port
6. method
7. errorcode

Then you're sending network packets with this information in a raw binary format (not a standardized or documented format, that i can tell) to some arbitrary UDP port on some arbitrary ipaddress.

Ignoring the question of "Why would you want that?", i have two direct objections to this.

1. The packet should be formatted according to some documented format somewhere. I suppose I don't really care all that much what the format is, but something that can be parsed with more than just "The first field is this many bytes, the second field is this many bytes, etc". Otherwise any hope of evolving the packet format in the future is lost.
2. UDP packets are not suitable for guaranteed delivery, so you're very likely to lose lots of log statements.

--------------------------------------

I have a broader, higher level, feedback consideration here as well.

Today this is only trying to provide notification support for a single message type.

But in general this looks very much like a generic logging mechanism that just happens to only be used for a single log type, why limit to only one, and why roll your own? Surely it would be more straight forward to replace the existing log macros with whatever logging implementation you want (potentially including this notification framework you've built)? If you don't like the current logging interface, this change replaces it with a third party open source logging library, that (I haven't double checked) probably has hooks to handle custom log sinks: https://github.com/coturn/coturn/pull/1351

--------------------------------------

Lastly, the dreaded "Why not X" question.

Why not use a well established telemetry system? That's basically what you're looking for right? A way to receive notifications on a separate process / system that can audit commands sent into `coturn` right?

I recommend instead of implementing an in-house solution for this, use an industry standard.

One that my company has been putting a lot of work into adopting recently is OpenTelemetry: https://github.com/open-telemetry/opentelemetry-cpp

It provides lots of capabilities for auditing, and the data is vendor / backend agnostic, with the ability to target quite a few existing vendors, or to create your own receiver.

Issue #1293, opened 2023-10-11
Hash algorithm that will be used in REST API setup


alexchow94:
Hello,
What will be the HMAC algorithm used by coturn? Is there a way to specify/set this in the config?
For eg: If the client sends the password with HMAC sha-256, will coturn be able to understand or know to use the same algo to generate its own HMAC to compare and validate?

alexchow94:
Disregard this issue.
Understood the concept of HMAC and how the client/server will validate the auth.
Closing this.
Thank you.

alexchow94:
Sorry, re-opening this issue.
Still couldn't find the info on how coturn understand/know to use the same algo to generate its own HMAC for comparison and validate.
Kindly help.
Thanks.

alexchow94:
Hello,
I found out coturn uses only HMAC-sha1, will there be support for >SHA256 in the near future?

alexg-axis:
I've opened a really quick and dirty PoC here (#1447). I'm pretty sure I'm missing something. But it does work in practice with the peers I tested with. So the question is more or less, will this break compatibility when changed or is the token/hmac just used by coturn and whatever produced the token in the first place?

GiantKs:
> I've opened a really quick and dirty PoC here (#1447). I'm pretty sure I'm missing something. But it does work in practice with the peers I tested with. So the question is more or less, will this break compatibility when changed or is the token/hmac just used by coturn and whatever produced the token in the first place?

ohh, let me know when it's available or if there's any progress...

PR #1447, opened 2024-03-06
Make REST API hash algorithm configurable

alexg-axis:
Add support for configuring the hash algorithm used for the "REST API
tokens" configurable. This makes it possible to use SHA-256 rather than
SHA-1, for example.

Solves #1293

### DIFF
--- a/src/apps/relay/userdb.c
+++ b/src/apps/relay/userdb.c
@@ -552,20 +552,19 @@ int get_user_key(int in_oauth, int *out_oauth, int *max_session_time, uint8_t *u
 
       hmac[0] = 0;
 
-      stun_attr_ref sar = stun_attr_get_first_by_type_str(
-          ioa_network_buffer_data(nbh), ioa_network_buffer_get_size(nbh), STUN_ATTRIBUTE_MESSAGE_INTEGRITY);
-      if (!sar) {
-        return -1;
-      }
-
-      int sarlen = stun_attr_get_len(sar);
-      switch (sarlen) {
-      case SHA1SIZEBYTES:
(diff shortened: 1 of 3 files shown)


src/apps/relay/userdb.c
```diff
@@ -552,20 +552,19 @@ int get_user_key(int in_oauth, int *out_oauth, int *max_session_time, uint8_t *u
 
       hmac[0] = 0;
 
-      stun_attr_ref sar = stun_attr_get_first_by_type_str(
-          ioa_network_buffer_data(nbh), ioa_network_buffer_get_size(nbh), STUN_ATTRIBUTE_MESSAGE_INTEGRITY);
-      if (!sar) {
-        return -1;
```
jonesmz, line 558: this is changing the behavior from using the packet to drive the hash choice, to using the commandline param / config file param.

I'm not sure that changing this behavior will be backwards compatible with existing users.

Can you elaborate on the compatibility considerations for this?

alexg-axis: It was my understanding that the token is not controlled by the specification for TURN and that the hash choice being based on the packet was a decision made by coturn developers. As coturn currently only seems to support [SHA-1 for integrity of the messages as per spec](https://github.com/coturn/coturn/blob/412788b120a8db0c9f9bc998b88892de39eb0ce7/README.md?plain=1#L118), I thought defaulting to SHA-1 would keep the existing behavior unless the command-line parameter is used. Please see the discussion in #1293. I'm not sure this is the right way to do it. So please clarify any assumptions I might be wrong about.

PR #1463, opened 2024-03-29
generate_enc_password: increase rsalt by 2

steveej:
before this change i see a bufferflow during `readable_string`.

### DIFF
--- a/src/client/ns_turn_msg.c
+++ b/src/client/ns_turn_msg.c
@@ -292,8 +292,9 @@ static void readable_string(unsigned char *orig, unsigned char *out, size_t sz) 
   out[0] = 0;
 
   for (i = 0; i < sz; ++i) {
-    snprintf((char *)(out + (i * 2)), 4, "%02x", (unsigned int)orig[i]);
+    snprintf((char *)(out + (i * 2)), 3, "%02x", (unsigned int)orig[i]);
   }
+  out[sz * 2] = 0;
 }
 
 static void generate_enc_password(const char *pwd, char *result, const unsigned char *orig_salt) {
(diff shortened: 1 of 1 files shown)


ggarber:
Thank you very much for the report @steveej and the fix! 🙇  
I think the problem is in the readable_string function and we shouldn't workaround it here, but let me take a look.

ggarber:
@steveej Could you try this alternative fix?   Thank you!

static void readable_string(unsigned char *orig, unsigned char *out, size_t sz) {
  size_t i = 0;
  out[0] = 0;

  for (i = 0; i < sz; ++i) {
    snprintf((char *)(out + (i * 2)), 3, "%02x", (unsigned int)orig[i]);
  }
  out[sz * 2] = 0;
}

src/client/ns_turn_msg_defs.h
```diff
@@ -60,9 +60,9 @@
 #define GET_STUN_ERR_RESP(msg_type) (msg_type | 0x0110)
 
 /* Lifetimes: */
-#define STUN_DEFAULT_ALLOCATE_LIFETIME (600)
+#define STUN_DEFAULT_ALLOCATE_LIFETIME (20)
```
eakraly, line 63: Were these changed for debugging?
Those should be reverted

eakraly (changes requested):

ggarber:
@steveej Can you take a look at the comments in this PR when you have a minute? Thank you very much !

src/client/ns_turn_msg_defs.h
steveej: thanks for catching this, those commits weren't supposed to be in this PR. i took them out however i still have other suggestions to consider.

steveej:
hey @ggarber all comments have been addressed

ggarber:
Thank you very much @steveej !

Issue #756, opened 2021-04-27
Add IP addresses in log messages


Cacodaimon:
For me as an admin I would like to see the IP addresses in the log messages in order to block suspicious connection¹ by utilizing a intrusion prevention system like fail2ban.

¹ As mentioned here: https://github.com/coturn/coturn/issues/737, https://github.com/coturn/coturn/issues/626

misi:
Source IP addresses of all incoming STUN or TURN requests? 
STUN requests is not authenticated so design logging of coturn in production needs to be prepared on rate limiting to avoid DoS.

Cacodaimon:
Thanks for your response!

> Source IP addresses of all incoming STUN or TURN requests?

The IP of incoming TURN connections would satisfy my needs to block unauthenticated connections.

> STUN requests is not authenticated so design logging of coturn in production needs to be prepared on rate limiting to avoid DoS.

This is my rsyslog filter `/etc/rsyslog.d/22-coturn.conf` currently used to avoid DOS.

```
# Syslog: coturn disable bind log spam
:msg, contains, "user <>: incoming packet BINDING processed, success" stop
```

But hence I use TURN only with authentication I would rather block those IPs.

ptroms:
iptables may help:
`sudo iptables -t raw -A PREROUTING -p tcp --dport 3478:3479 -j TRACE`
or:
`sudo iptables -t raw -A PREROUTING -p tcp --dport 3478:3479 -j TRACE`
or whatever ports your serv listening on. It ends up in syslog, and you can pickup does IP's form there.

Cacodaimon:
@ptroms Thanks for the advice but in this case I get only the IP and not the information whether the user is authenticated or not.

I want to block the IPs doing this kind of requests:

"session 000000000000000002: realm <signalling.example.de> user <>: incoming packet message processed, error 401: Unauthorized"
"session 000000000000000001: realm <signalling.example.de> user <123456789:turn-test-user>: incoming packet message processed, error 401: Unauthorized"

tgrushka:
Why can't Coturn have at least a config option to add client IP addresses to logs? That's pretty basic and almost every other server out there has this ability.

Oklit:
+1 as this is really annoying that there is no easy option to check who is connecting (yes - I also would like to use fail2ban to block some requests).

Stoic-Lithium:
I also desire for IP addresses to be included in the log output.

Talkless:
I see scanners doing:

```
[...]

PR #1481, opened 2024-05-02
Add remote ip to logs

obydog002:
Address https://github.com/coturn/coturn/issues/756

Add remote ips to logging

### DIFF
--- a/src/server/ns_turn_server.c
+++ b/src/server/ns_turn_server.c
@@ -81,6 +81,10 @@ int TURN_MAX_ALLOCATE_TIMEOUT = 60;
 int TURN_MAX_ALLOCATE_TIMEOUT_STUN_ONLY = 3;
 
 static inline void log_method(ts_ur_super_session *ss, const char *method, int err_code, const uint8_t *reason) {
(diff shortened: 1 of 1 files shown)


ggarber:
Thank you very much @obydog002.  Can you review the lint errors?  Maybe it is because the lines are too long now?
(ignore the msvc-analyzer failures, those have been there for a while I think)

obydog002:
@ggarber  :)

obydog002:
can we merge this?

jelmd:
IMHO this change adds a lot of overhead and thus should not be merged. Logging the IPs when a session gets established is sufficient and is already a DoS vector (an admin should be able to turn it on/off on demand). Logging IPs for each incoming packet makes it worse and hurts performance. Also note, that logging IPs could be a problem wrt. GDPR compliance ...

eakraly:
There is a new mechanism to rate limit invalid traffic enable through `--unauthorized-ratelimit`
Please give it a try
Logging invalid IPs and then blocking them through firewall might be counter productive as it will reduce performance of the server and increase impact of attacks

PR #1488, opened 2024-05-13
Allow authenticating with a username to redis

maddymeows:
I needed this

I just looked for occurrences of password and added username branches there. Let me know if there's anything wrong, I'm not proficient in C at all.

### DIFF
--- a/src/apps/relay/dbdrivers/dbd_redis.c
+++ b/src/apps/relay/dbdrivers/dbd_redis.c
@@ -49,6 +49,7 @@ static void turnFreeRedisReply(void *reply) {
 struct _Ryconninfo {
   char *host;
   char *dbname;
+  char *user;
   char *password;
   unsigned int connect_timeout;
   unsigned int port;
@@ -63,6 +64,9 @@ static void RyconninfoFree(Ryconninfo *co) {
     }
     if (co->dbname) {
       free(co->dbname);
+    }
+    if (co->user) {
+      free(co->user);
     }
     if (co->password) {
       free(co->password);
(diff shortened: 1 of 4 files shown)


eakraly:
Thank you @maddymeows !
It seems it would allow coturn to work with Redis6 that has new ACL feature

Could you please rebase the PR and also explain if you tested with new and old redis.
Does it require minimal version of hiredis to support it?

maddymeows:
Sorry for tagging the regression fix alongside this PR, came up in testing after rebasing.

> Could you please rebase the PR
Done

> and also explain if you tested with new and old redis.
I've tested on debian 10 which has redis dated before ACLs were introduced, works fine as long as you don't configure a username.
I've also tested on latest debian and fedora and works just as well

> Does it require minimal version of hiredis to support it?
As far as I know, no

src/apps/relay/dbdrivers/dbd_redis.c
```diff
@@ -163,9 +167,6 @@ static Ryconninfo *RyconninfoParse(const char *userdb, char **errmsg) {
     if (!(co->host)) {
       co->host = strdup("127.0.0.1");
     }
-    if (!(co->password)) {
```
eakraly, line 166: Why this change?

maddymeows: Originally during testing I created an extra branch `if (!(co->user)) { co->user = strdup(""); }`.

However, I found out that it made coturn always authenticate with a username and password, which would break for redis servers that don't have the ACL feature.

That's because the branches using password checked if the pointer to password was set, which it always was with the statement I just removed, rather than if password contained a useful value. So the password conditionals haven't been working for a while, coturn just discarded the error that redis threw out and continued on (which is fine, but not ideal).

Hence also changes to make those branches `co->password && strlen(co->password)` rather than just `co->password`, which I could effectively change to `co->password && co->password[0]` but strlen felt nicer.

Error handling of missing auth doesn't seem to be handled fully correctly, but happens moreso as a side effect, the side effect of using the side effect is it never warns about if the password is needed which is why there are no logs complaining about use of `AUTH` when wasn't needed.

src/apps/relay/dbdrivers/dbd_redis.c
```diff
@@ -268,7 +273,7 @@
         }
       }
 
-      ret = redisLibeventAttach(base, co->host, co->port, co->password, atoi(co->dbname));
+      ret = redisLibeventAttach(base, co->host, co->port, co->user, co->password, atoi(co->dbname));
```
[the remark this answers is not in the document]
maddymeows: `co->dbname` is never null because a default is given in `RyconninfoParse`, I don't know how to silence this.

eakraly:
Hi @maddymeows ! Could you please update the PR so I can merge it into upcoming version? Thanks

maddymeows:
@eakraly done!

justinschoeman:
This commit breaks unauthenticated redis connections ("select <dbname>" is not sent if password is not set).

maddymeows:
> This commit breaks unauthenticated redis connections ("select " is not sent if password is not set).

<https://github.com/coturn/coturn/pull/1668>

PR #1491, opened 2024-05-21
Fix mingw and MSVC ci build

jonesmz:


src/apps/common/apputils.c
```diff
@@ -355,7 +359,7 @@ int get_raw_socket_ttl(evutil_socket_t fd, int family) {
     } while (0);
 #else
     socklen_t slen = (socklen_t)sizeof(ttl);
-    if (getsockopt(fd, IPPROTO_IPV6, IPV6_UNICAST_HOPS, &ttl, &slen) < 0) {
+    if (getsockopt(fd, IPPROTO_IPV6, IPV6_UNICAST_HOPS, (char *)&ttl, &slen) < 0) {
```
eakraly, line 362: ttl is defined as int - should it be redefined as char then?
Otherwise we may get wrong result - if value is put in the wrong byte of the int (not sure if this scenario is possible but why having this complication in the first place)

The other option is to make it `(int*)&ttl`

src/apps/common/apputils.c
```diff
@@ -391,7 +395,7 @@ int get_raw_socket_tos(evutil_socket_t fd, int family) {
     } while (0);
 #else
     socklen_t slen = (socklen_t)sizeof(tos);
-    if (getsockopt(fd, IPPROTO_IPV6, IPV6_TCLASS, &tos, &slen) < 0) {
+    if (getsockopt(fd, IPPROTO_IPV6, IPV6_TCLASS, (char *)&tos, &slen) < 0) {
```
eakraly, line 398: Same comment as for ttl (char vs int)

src/apps/common/apputils.c
```diff
@@ -704,10 +708,10 @@ int get_socket_mtu(evutil_socket_t fd, int family, int verbose) {
   int val = 0;
   socklen_t slen = sizeof(val);
   if (family == AF_INET) {
-    ret = getsockopt(fd, IPPROTO_IP, IP_MTU, &val, &slen);
+    ret = getsockopt(fd, IPPROTO_IP, IP_MTU, (char *)&val, &slen);
```
eakraly, line 711: char vs int

eakraly (changes requested):

src/apps/common/apputils.c
jonesmz: Well `&ttl` is already an `int*`, so casting it like that wouldn't do anything.

Microsoft's documentation for this function ( https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options ) says that the argument is DWORD (aka, int)

Linux man page ( https://man7.org/linux/man-pages/man7/ipv6.7.html ) says 

> Argument is a pointer to an integer. -1 in the value means use the route default, otherwise it should be between 0 and 255.

On windows, `getsockopt` explicitly takes argument of `char*` : https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-getsockopt

On POSIX (aka, linux's implementation), it's `void*` https://pubs.opengroup.org/onlinepubs/009604599/functions/getsockopt.html

The MinGW compiler explicitly uses the Microsoft API, not the POSIX one, so it's trying to use the form that takes `char*`.

```
C:/msys64/mingw64/include/winsock2.h:1010:82: note: expected 'char *' but argument is of type 'int *'
 1010 |   WINSOCK_API_LINKAGE int WSAAPI getsockopt(SOCKET s,int level,int optname,char *optval,int *optlen);
```

So my understanding of all of this is that `ttl` should be of type `int`, and casting it to `char*` is correct to support compiling on the Windows platform.

I could try casting to `void*` but i'm not sure that'll implicitly convert to `char*` on MinGW?

jonesmz: > why having this complication in the first place

Yeeeeeeep, POSIX API is... yea.

eakraly:
Assuming is it messy and workarounds will only make it messier... Let's keep it this way
But we should at least ensure it works. Is there a way you can confirm these changes work as intended @jonesmz ?
Thanks!

jonesmz:
> Assuming is it messy and workarounds will only make it messier... Let's keep it this way But we should at least ensure it works. Is there a way you can confirm these changes work as intended @jonesmz ? Thanks!

Well, I do have the latest version using `(void*)` instead of `(char*)` apparently mingw auto converts it.

As for being able to test it... not easily, and not for several weeks. I'm currently on parental leave from work, so I couldn't post this version to our test environment to do basic tests there until I return in a month or so.

It'd be better to rely on unit tests for that kind of thing, but I imagine that's hard to do anything about when there already aren't very many.

jonesmz:
If you want, I can break the MSVC commit into it's own PR. Obviously the mingw changes are more involved.

eakraly:
> As for being able to test it... not easily, and not for several weeks. I'm currently on parental leave from work, so I couldn't post this version to our test environment to do basic tests there until I return in a month or so.
> 
Congrats! and thank you for contributing here while working on +1 to your family :)

jonesmz:
> > As for being able to test it... not easily, and not for several weeks. I'm currently on parental leave from work, so I couldn't post this version to our test environment to do basic tests there until I return in a month or so.
> 
> Congrats! and thank you for contributing here while working on +1 to your family :)

The wife has me on "night shift" to feed and change the newborn while she's sleeping. Gives me a few hours a night with nothing to do other than dispense bottles and diapers ;)

PR #1495, opened 2024-05-22
Check the result of malloc in string_list_add

jonesmz:


src/server/ns_turn_maps.c
```diff
@@ -993,9 +993,16 @@ static string_list *string_list_add(string_list *sl, const ur_string_map_key_typ
     return sl;
   }
   string_elem *elem = (string_elem *)malloc(sizeof(string_elem));
+  if (!elem) {
```
eakraly, line 996: I do not know if these kind of changes are good actually

I have never seen any application successfully run past failure of allocation of memory except for managed languages

But assuming it does run, this will create silent state corruptions with hard to understand implications: we tried to add a new value to map but failed and returned
This method's promise is that the value is ALWAYS added as long as the key is not null - this PR breaks this promise silently.

eakraly (changes requested):

src/server/ns_turn_maps.c
jonesmz: It's very easy to create out of memory conditions on windows or embedded linux if you go outside of the default settings.

E.g. you can put a hard limit on the number of bytes a process is allowed to allocate using cgroups on linux, and then any allocations that try to go past that will return null.

This is this code scanning report: https://github.com/jonesmz/coturn/security/code-scanning/26

If coturn is running in an environment where malloc won't fail, and the OOM killer will just kill the process, then checking the result of malloc doesn't hurt and we can't get into an inconsistent state.

If it is running in an environment where malloc can fail, then it's probable that there'll be a state inconsistency with the way the rest of the code is currently written... but that can and should be addressed by additional patches. However, we can't get from here to there without making some improvements and addressing the security scanning results is an important preliminary step to get the mess under control.

jonesmz: Elaborating a bit more:

I'd really like to get the code scanners satisfied. From my perspective, they're low hanging fruit for code quality, and addressing them allows for us to get a baseline where PRs can be required to keep the number of scanning results at 0 without anyone having to think about it.

eakraly: I agree on all points
What I would add is extend/change this API to clearly signal that new element was not added
In this case we will not have to chase all the scenarios where those silent failures occured

What I am trying to avoid is adding those silent failures by: explicitly crashing, logging (though it will not work because OOM), returning or having output argument that operation failed etc.

jonesmz: I don't think that this change in "fails because allocation failed" is user visible in a meaningful way.

If the code is running in an environment where allocation fails without invoking OOMkiller, then the code is already riddled with situations where that'll break in weird ways. Many of them, of course, will get a nullptr dereference right away but not all of them.

If the code isn't running in a constrained environment like that, then malloc will never return null and the OOMkiller will run and this change to check for nullptr is not visible.

My preference is that `coturn` simply say "Running in an environment where allocations can fail (e.g. malloc returns null) is not supported, and you're on your own"

Then later, after the security scanner is happy, and we've adopted the ability to signal allocation failures, and all the code-paths which might need to react to that have been updated to react to them, then we can say "Running in an environment where allocations can fail is supported".

Basically, I don't disagree with you, I just don't think that that improvement should delay merging PRs like this.

jonesmz: Especially since 

> ability to signal allocation failures, and all the code-paths which might need to react to that have been updated to react to them

Is basically C++ exceptions with extra steps, so by the time I have the energy and willingness to address allocation failure safety, I'll already be well on my way to converting more of the code to fluent C++.

eakraly (approved):

PR #1499, opened 2024-05-23
Avoid read-past-end of string in get_bold_admin_title

jonesmz:


eakraly:
@jonesmz It does not really read past end of string
It copies up to 1025 chars to sbat(which has size of 1025) so no "past end of string" is happening here

What you did is "copy 1024 and set char 1025 to '/0' to make it null terminated string if __bold_admin_title is longer"

While your change is correct it does not do anything
Additional check of `as` in if is actually good one (but is not reflected in the name of PR)

jonesmz:
> It copies up to 1025 chars to sbat(which has size of 1025) so no "past end of string" is happening here
>
> What you did is "copy 1024 and set char 1025 to '/0' to make it null terminated string if __bold_admin_title is longer"

```
strncpy(sbat, __bold_admin_title, sizeof(sbat));
```

If the variable `__bold_admin_title` is 1025 bytes or longer, than `sbat` is not nul-terminated by `strncpy()`

then any calls to `strlen(sbat);` or any places where the return value from the function (which is the pointer to `sbat`) is used (presumably via `strlen()`) will read past the end.

This PR addresses this security scanning result: https://github.com/jonesmz/coturn/security/code-scanning/79

but i'm not sure anyone but my account can read that. There should be an equivelent report in the `coturn` repo's security tab (but I can't read that since I'm not part of the coturn project's collaborator list).

eakraly:
> > It copies up to 1025 chars to sbat(which has size of 1025) so no "past end of string" is happening here
> > What you did is "copy 1024 and set char 1025 to '/0' to make it null terminated string if __bold_admin_title is longer"
> 
> ```
> strncpy(sbat, __bold_admin_title, sizeof(sbat));
> ```
> 
> If the variable `__bold_admin_title` is 1025 bytes or longer, then `sbat` is not nul-terminated by `strncpy()`
> 
> From there, any calls to `strlen(sbat);` or any places where the return value from the function (which is the pointer to `sbat`) is used (presumably via `strlen()`) will read past the end.
> 
> This PR addresses this security scanning result: [jonesmz/coturn/security/code-scanning/79](https://github.com/jonesmz/coturn/security/code-scanning/79)
> 
> but i'm not sure anyone but my account can read that. There should be an equivelent report in the `coturn` repo's security tab (but I can't read that since I'm not part of the coturn project's collaborator list).

Yes, setting last char of `sbat` to null makes sense though the initial string we copy into it is our string and not externally provided so the chance of it overflowing is zero.
In any case, I think the difference between our point of views is negligible.

eakraly (approved):

PR #1500, opened 2024-05-23
Avoid writing potentially uninitialized data to aes_128 key file

jonesmz:


src/apps/relay/mainrelay.c
```diff
@@ -1740,32 +1740,40 @@ void encrypt_aes_128(unsigned char *in, const unsigned char *mykey) {
...
+    return;
+  }
+  for (size_t i = 0; i < 16; ++i) {
     fputc(key[i], fptr);
   }
-  STRCPY((char *)returnedKey, key);
+  memcpy(returnedKey, key, 16);
+  // Note: Don't put a nul-terminator at the end.
+  // this function is only ever called with returnedKey
+  // as fixed size char arrays of size 16.
```
jonesmz, line 1776: STRCPY is a very dangerous macro

```
#defineSTRCPY(dst, src)                                                                                               \
  do {                                                                                                                 \
    if ((const char *)(dst) != (const char *)(src)) {                                                                  \
      if (sizeof(dst) == sizeof(char *))                                                                               \
        strcpy(((char *)(dst)), (const char *)(src));                                                                  \
      else {                                                                                                           \
        size_t szdst = sizeof((dst));                                                                                  \
        strncpy((char *)(dst), (const char *)(src), szdst);                                                            \
        ((char *)(dst))[szdst - 1] = 0;                                                                                \
      }                                                                                                                \
    }                                                                                                                  \
  } while (0)
```
when src is a fixed length char array that isn't guaranteed to be nul-terminated, it'll read past the end of src.

Obviously so will the strcpy function.

In general, the string handling in c-language is considered (and just is) a massive security cluster-f for a reason.

Switching to memcpy here is semantically correct, because key is a fixed length char array with all indices initialized, and no nul-terminator.

but because we don't know the size of the buffer provided, we can't know if writing a nul-terminator is safe or not... so we have to look at all the places where the function is called to determine that it's always called with a fixed size 16 char array.

src/apps/relay/mainrelay.c
```diff
@@ -1740,32 +1740,40 @@ void encrypt_aes_128(unsigned char *in, const unsigned char *mykey) {
   unsigned char *base64_encoded = base64encode(total, totalSize);
   printf("%s\n", base64_encoded);
 }
-void generate_aes_128_key(char *filePath, unsigned char *returnedKey) {
-  int i;
-  int part;
-  FILE *fptr;
+static void generate_aes_128_key(char *filePath, unsigned char *returnedKey) {
```
jonesmz, line 1743: function is only called from this translation unit. Making it static will help the analyzer understand usage.

src/apps/relay/mainrelay.c
eakraly: Yes, STRCPY  is on of the dangerous macros in the code

eakraly:
@jonesmz thank you for the contribution!
Seems like another `static` key word is required in the declaration

PR #1501, opened 2024-05-23
Check allocation results in add_static_user_account

jonesmz:


src/apps/relay/userdb.c
```diff
@@ -717,52 +717,78 @@ void release_allocation_quota(uint8_t *user, int oauth, uint8_t *realm) {
...
+  // the ur_string_map functions only fail (well... other than allocation failures, which aren't handled)
+  // if the map isn't valid. So we only need to check the result of locking.
+  if (ur_string_map_lock(turn_params.default_users_db.ram_db.static_accounts) < 0) {
+    free(usname);
+    free(key);
+    return -1;
   }
 
-  return -1;
+  // key argument is deep-copied, so ownership isn't transfered, and we still need to free usname later..
```
eakraly, line 782: If key is deep-copied then we should also free it at the end, no?

jonesmz: so the parameter "usname" is the key, and the parameter "key" is the value, so we've got a terminology mismatch here.

I suppose i should more carefully word the comment.

eakraly (changes requested):
You replaced calloc with malloc but did not update arguments (which fails build)

jonesmz:
> You replaced calloc with malloc but did not update arguments (which fails build)

Right, my mistake there. Thanks!

eakraly (approved):

PR #1502, opened 2024-05-23
Change the various map functions to return bool instead of inconsistantly return 0, 1, or -1

jonesmz:


eakraly:
Thank you @jonesmz - this is a great change!

eakraly (approved):

PR #1507, opened 2024-05-28
malloc now allocates space for string terminator

redraincatching:
addresses a code scanner vulnerability

the combination of `strlen` and `malloc` results in space being allocated for the string, but not the null terminator required to end the string, so space for an extra character has to be manually specified

#### references
- CERT C Coding Standard: [MEM35-C. Allocate sufficient memory for an object](https://www.securecoding.cert.org/confluence/display/c/MEM35-C.+Allocate+sufficient+memory+for+an+object).
- Common Weakness Enumeration: [CWE-131](https://cwe.mitre.org/data/definitions/131.html).
- Common Weakness Enumeration: [CWE-120](https://cwe.mitre.org/data/definitions/120.html).
- Common Weakness Enumeration: [CWE-122](https://cwe.mitre.org/data/definitions/122.html).

### DIFF
--- a/src/apps/relay/dbdrivers/dbd_mysql.c
+++ b/src/apps/relay/dbdrivers/dbd_mysql.c
@@ -115,7 +115,7 @@ char *decryptPassword(char *in, const unsigned char *mykey) {
 #endif
 
   strcat(last, (char *)outdata);
-  out = (char *)malloc(sizeof(char) * strlen(last));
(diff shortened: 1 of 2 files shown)


eakraly:
Thank you @redraincatching !
That whole function is not safe - there is no check on size of the input or if the output will fit into the intermediary buffer. There are also 2 intermediary buffers (one would have been enough)...
There is a lot more to do :)

eakraly (approved):

PR #1508, opened 2024-05-30
Don't call `setgroups` unconditionally in mainrelay

networkException:
This pull request moves the call to `setgroups` from the beginning of the `drop_priviliges` function to branch in which `setuid` is actually called. This still fulfills the intention of acbf7e15c9290e0891a6b6b5ce6e81bbaa77ce5a, initially introducting the call to `setgroups`:

> Fix related to POS36-C and rpmlint error "missing-call-to-setgroups-before-setuid".

As per this intention is is not required to call `setgroups` otherwise, reducing the more exotic (as in not part of POSIX and considered priviliged by systemd) system calls coturn needs to make at startup.

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -2913,7 +2913,6 @@ static void drop_privileges(void) {
 #if defined(WINDOWS)
   // TODO: implement it!!!
 #else
-  setgroups(0, NULL);
   if (procgroupid_set) {
     if (getgid() != procgroupid) {
       if (setgid(procgroupid) != 0) {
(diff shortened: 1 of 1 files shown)


eakraly:
Thank you @networkException for the contribution?
Is there any way we can test it?

networkException:
You can run

```
systemd-run -t -p "SystemCallFilter=~@privileged" turnserver
```

and see the process get killed without this patch (`journalctl -t kernel -f` for the error message)

I don't have a setup to test if rpmlint accepts this sadly

mweinelt:
I have been running with this patch and `~@privileged` since last October. I'm also configuring explicit listening addresses.

mweinelt:
We have been shipping this in nixpkgs since mid 2025, can this be merged?

PR #1509, opened 2024-05-30
Move the hiredis_libevent2 code from common to relay

jonesmz:
The point of this change is to make the build instructions a bit more straight forward. Since the hiresevent2 source files are only ever used by the relay target, this scoping makes sense in general.


jonesmz:
I also clean up a bunch of superfluous whitespace while i'm at it.

eakraly (approved):

eakraly:
nice!

PR #1510, opened 2024-05-30
Add spdx tags to all source files

jonesmz:
With notable exceptions of:

src/apps/common/win/*
src/apps/relay/telnet.*


jonesmz:
@eakraly since this is a comments-only change, can we merge it even though the continuous-integration is currently broken?

Anything you need adjusted before merging?

jonesmz:
> @eakraly since this is a comments-only change, can we merge it even though the continuous-integration is currently broken?
> 
> Anything you need adjusted before merging?

Just checking if anything is needed before merging this.

jonesmz:
@eakraly I'm sorry to be a nag, but it's been 4 months since i opened this comment-only change.

Is the change objectionable in some way?

As I briefly discussed with you in previous reviews that included this commit, the change is desirable from a software provenance point of view, and the specific comment format is from the linux foundation. My work has automatic tools that verify license compliance and this would help us ensure that we comply properly. Similarly, other organizations that use similar automated tools would benefit in the same way.

PR #1527, opened 2024-06-25
windows: Only attempt to bind when the network interface is up

davids-work:
Previously if the system had an interface with a static IP configured, coturn would attempt to bind to that address, even if the interface was down. This would fail, and prevent coturn from starting (even if there were other usable interfaces)


eakraly:
Thank you @davids-work 
This makes sense - looks like non-windows version has a similar check already (`if (!(ifa->ifa_flags & IFF_UP))`)

PR #1529, opened 2024-06-25
Add new Drain feature

sgodin:
Add new Drain feature

-when coturn server is in drain mode
  -current allocations will continue to work as usual
  -new allocations will be rejected with a 403 (Forbidden) response
  -when all allocations go away, then coturn will shutdown
-Enable drain mode with either
  -signaling SIGUSR1
  -turn_admin_server "drain" CLI command

This contribution is from Wire. https://wire.com/


sgodin:
FYI - This contribution is from Wire.  https://wire.com/

src/apps/relay/mainrelay.c
```diff
@@ -188,7 +188,8 @@ turn_params_t turn_params = {
     {NULL, 0, {0, NULL}}, /*tls_alternate_servers_list*/
 
     /////////////// stop server ////////////////
-    0, /*stop_turn_server*/
+    DRAINMODE_NOT_ENALBED, /*drain_turn_server*/
```
jonesmz, line 191: Enalbed -> enabled

src/apps/relay/userdb.c
```diff
@@ -68,6 +68,9 @@ static TURN_MUTEX_DECLARE(o_to_realm_mutex);
 static ur_string_map *o_to_realm = NULL;
 static secrets_list_t realms_list;
 
+int global_allocation_count = 0; // used for drain mode, to know when all allocations have gone away
```
jonesmz, line 71: Why use a mutex protected into over an atomic int? An atomic into would be much less cumbersome to use and faster to access.

jonesmz: Also, use size_t over into for "count" of things.

src/server/ns_turn_server.h
```diff
@@ -203,6 +203,9 @@ struct _turn_turnserver {
   /* Return an HTTP 400 response to HTTP connections made to ports not
      otherwise handling HTTP. */
   vintp respond_http_unsupported;
+
+  /* Set to true on SIGUSR1 */
+  int is_draining;
```
jonesmz, line 208: This is a bool. Use the bool type from stdbool.h

jonesmz: And the true and false keywords from the same header in the places where you change the value

src/server/ns_turn_server.c
```diff
@@ -1294,6 +1294,12 @@ static int handle_turn_allocate(turn_turnserver *server, ts_ur_super_session *ss
           }
         }
 
+        if (server->is_draining) {
+          // Don't allow new allocations if we are draining
+          *err_code = 300; // 300 (Try Alternate): The client should contact an alternate server for this request.
```
jonesmz, line 1299: Try alternate error needs to carry an alternate server ipaddress or domain

https://datatracker.ietf.org/doc/html/rfc8489#section-10

src/apps/relay/netengine.c
```diff
@@ -1605,6 +1605,32 @@ void run_listener_server(struct listener_server *ls) {
...
+        for (int j = 0; j < get_real_general_relay_servers_number(); j++) {
+          if (ls->aux_udp_services[i] && ls->aux_udp_services[i][j]) {
+            ls->aux_udp_services[i][j]->ts->is_draining = 1;
+          }
+        }
+      }
+      turn_params.drain_turn_server = DRAINMODE_ENABLED;
+    } else if (turn_params.drain_turn_server == DRAINMODE_ENABLED && global_allocation_count == 0) {
+      turn_params.stop_turn_server = 1;
+      TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Drain complete, shutting down now...\n");
```
jonesmz, line 1631: Use true keyword

src/apps/relay/mainrelay.h
```diff
@@ -176,6 +176,8 @@ enum _NET_ENG_VERSION {
 
 typedef enum _NET_ENG_VERSION NET_ENG_VERSION;
 
+typedef enum _DRAINMODE_STATE { DRAINMODE_NOT_ENALBED = 0, DRAINMODE_REQUESTED = 1, DRAINMODE_ENABLED } DRAINMODE_STATE;
```
jonesmz, line 179: I'm not sure I follow why an enum is needed here.

We're either in drain mode or not, right?

A stare transition from "not draining" to "requested" to "draining" doesn't seem to be useful.

src/apps/relay/netengine.c
```diff
@@ -1605,6 +1605,32 @@ void run_listener_server(struct listener_server *ls) {
...
+          if (ls->dtls_services[i] && ls->dtls_services[i][j]) {
+            ls->dtls_services[i][j]->ts->is_draining = 1;
+          }
+        }
+      }
+      for (size_t i = 0; i < turn_params.aux_servers_list.size; i++) {
+        for (int j = 0; j < get_real_general_relay_servers_number(); j++) {
+          if (ls->aux_udp_services[i] && ls->aux_udp_services[i][j]) {
+            ls->aux_udp_services[i][j]->ts->is_draining = 1;
+          }
```
jonesmz, line 1625: Don't we need to also drain TCP and TLS?

src/apps/relay/mainrelay.c
sgodin: Thank you for reviewing so quickly!!! Nice find will fix it up.

src/apps/relay/userdb.c
sgodin: I wasn't sure that there was existing an atomic increment in C that all your supported compilers would be happy with.  Do you know what might supported by all OSs?

Good idea for size_t change, will do.

src/server/ns_turn_server.h
sgodin: I was just following suit with the rest of the code here, bool is not used at all in this struct, see members: fingerprint, rfc5780, external_ip_set, etc.  Are you sure you want me to change this and be different than other members?  I'm happy to change it, if you prefer.

jonesmz: My team at work is actively working on converting coturn to use stdbool.h, just slow going due to other things suddenly popping up.

It'd be nice to have fewer places to fix :-)

src/apps/relay/userdb.c
jonesmz: Shouldn't the C11 atomic functions work?

https://en.cppreference.com/w/c/atomic

I don't know of an official list of supported platforms for coturn. I'm not a maintainer.

That said, personally, if it's not in the CI system, i don't consider it a supported platform. Someone who cares about a platform that doesn't (somehow...?) have C11 atomics missed their chance to get a CI runner / action setup to ensure their needs are met.

They can always use older releases.

src/server/ns_turn_server.c
jonesmz: Well, I say "needs" but i guess i'm not 100% sure on that. Could be that you don't *strictly* speaking need it.

But if any alternatives are configured in the config file, probably those should be included in the message?

src/server/ns_turn_server.h
sgodin: While I agree, I also think that code consistency is important, it's confusing if the same thing is done 2 different ways.

src/server/ns_turn_server.c
jonesmz: Well the "USE-ALTERNATE" flag is supposed to be the mechanism for the turnserver to say "I can't service your request, but these other servers probably could".

Like you, my use of coturn involves DNS lookup, not raw ipaddresses, so i find it really annoying that the only use of ALTERNATE-DOMAIN attribute is worded as being only allowed for DTLS / TLS connections, when it realistically should work for any type of connection.

My concern with 403 forbidden is that the server might be construed as saying "You are not authorized, bad credentials" even though that's not actually the correct one.

Insufficient capacity has only the problem of the client potentially thinking that the server is unhappy about ports ( even-port / reservation-token ), therefore it may burn an additional round-trip trying to talk to the server before it gives up and moves on.

Unfortunately we can't invent our own error codes, as they would not be comprehended by clients.

jonesmz: note that the alternate server configuration option already exists

https://github.com/coturn/coturn/blob/fbe07c4a16f4e07d3ad199674016868657c4b29b/examples/etc/turnserver.conf#L574-L611

so you'd only need to read from that and populate, according to the socket type.

src/server/ns_turn_server.h
jonesmz: It's really very confusing to people to have ints used as bools.

Like, i can't make you, but please consider using bool, i'm just going to change it to bool later (and i'll do it one at a time...).

src/server/ns_turn_server.c
sgodin: RFC8489 - Does use the term MUST:

The server MUST include at least
   one ALTERNATE-SERVER attribute in the error response, which MUST
   contain an IP address of the same address family as the source IP
   address of the request message.

So I could just add the configured value, but I'm wondering if we might be better off with a different error code.  Any other opinions?

src/apps/relay/netengine.c
sgodin: This was another following suit item, see turn_admin_server.c:
turn_params.stop_turn_server = 1;

Happy to change it, if consistency is not important.

src/apps/relay/mainrelay.h
sgodin: The current process is staged.  

Step 1:  DRAINMODE_REQUESTED is flagged for the run_listener_server method to pick up
Step 2:  run_listener_server loop see's REQUESTED flag, then we turn on is_draining flag in each turn server.  Then drain mode transitions to DRAINMODE_ENABLED.

If I can get rid of the 2 staged approach, I won't need an enum here.  :)

src/apps/relay/mainrelay.c
```diff
@@ -3983,4 +3987,11 @@ static void shutdown_handler(evutil_socket_t sock, short events, void *args) {
   UNUSED_ARG(args);
 }
 
+static void drain_handler(evutil_socket_t sock, short events, void *args) {
+  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Draining then terminating on signal %d\n", sock);
+  turn_params.drain_turn_server = DRAINMODE_REQUESTED;
+
+  UNUSED_ARG(events);
```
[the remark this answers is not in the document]
sgodin: I was following suit with how the stop_turn_server flag was implemented. At the time I didn't think mainrelay.c had access to listener_server struct, but I see now that it is available via turn_params.listener.  

I will look into this further.  Then I can get rid of the enum, and turn_params.drain_turn_server.

src/apps/relay/netengine.c
sgodin: Yes, I suppose we do, I will look into this as well.

src/apps/relay/userdb.c
sgodin: I can try it out and see how the CI builds go.

src/server/ns_turn_server.c
sgodin: I will add the alternate server from configuration to the 300 response.

src/server/ns_turn_server.h
sgodin: Project maintainers do you have an opinion here?  Consistency vs better typing for new code?  

I really am happy to go either way. :)

src/server/ns_turn_server.c
sgodin: Turns out that's not really possible.  Right now the alternate-server configuration option enables a "redirection" mode, and all Allocate requests to this coturn instance get 300 responses.  I'm thinking it will be best stay away from the 300 response then.  I'm thinking to go with the 403 response - it's description sounds appropriate and it is in a TURN RFC.  It should not be mistaken for auth errors according to the description, 401's are for that.

src/apps/relay/netengine.c
sgodin: Actually I found a better way to do this, that means I don't need to move dtls_listener_relay_server_info to the header file.  Changes coming soon.

src/apps/relay/userdb.c
sgodin: Well it turns out the windows CI builds are not happy with stdatomic.h use.  I get this error when trying to use it in my VS2022:
The contents of <stdatomic.h> are available only with C++23 or later.

I will revert this back to mutex use tomorrow, unless someone prefers conditional code for windows vs non-windows.

jonesmz: My understanding of the visual studio compiler is that it is exclusively a c++ compiler with very minor affordances for c code.

You could use std::atomic, which my work uses in c++ on visual studio with no issue.

It has an interface much nicer than the interlocked functions.

src/apps/relay/userdb.c
```diff
@@ -685,6 +692,14 @@
     free(username);
     ur_string_map_unlock(rp->status.alloc_counters);
   }
+
+#ifndef _MSC_VER
+  global_allocation_count++;
+#else
+  InterlockedIncrement((volatile LONG *)&global_allocation_count);
+#endif
+  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Global turn allocation count incremented, now %ld\n", global_allocation_count);
```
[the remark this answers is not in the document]
sgodin: Only used in a log statement, I'm not worried about this.  Does it need to be addressed for this PR to get accepted?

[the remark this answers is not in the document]
jonesmz: I'm trying to get the static analyzer and security scanners to have 0 issues reported. 

We have quite the number of them.

I would appreciate it if you would resolve the complaint, but I can't stop the maintainers from merging as is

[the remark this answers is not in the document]
sgodin: okay should be fixed now

jonesmz (approved):
Aside from minor grumbling about stdbool.h and thinking std::atomic<size_t> would be a better fit on MSVC (i believe coturn always compiles as C++ on MSVC on purpose, but i'm not 100% sure about that) this seems like an excellent contribution.

sgodin:
Thanks for all your time looking at this, really appreciate it.  A number
of great changes came out of it.

As for the minor grumbling... :)
1.  I'm happy to change to use bool (vs the current implementation that is
consistent with the rest of the code) - if there is consensus from project
maintainers on this.
2.  I'm running out of patience today with the atomic stuff, just trying to
get the interlocked solution took me 7 or so commits, it is tedious to get
OS sensitive code right without reasonable access to all the various
compilers.  But I am happy to take another run at it, if needed for
acceptance.  :)

Thanks again for your time.

Scott

src/server/ns_turn_server.h
eakraly: A lot of code has been converted to bool already. Definitely much clearer code.

sgodin: okay - I will change to bool :)

sgodin:
Conversion to use bool is complete, I think this PR may be ready for acceptance at this point.  Thanks for all the feedback.

Scott

sgodin:
Hi Guys - just checking in.  Is there anything else you need changed on this PR for it to be accepted?  Thanks - I appreciate your time!

eakraly:
Sorry @sgodin somehow this slipped in between everything else
I'll try to review this weekend
Thank you for the contribution!

src/apps/relay/userdb.c
```diff
@@ -685,6 +692,15 @@ int check_new_allocation_quota(uint8_t *user, int oauth, uint8_t *realm) {
     free(username);
     ur_string_map_unlock(rp->status.alloc_counters);
   }
+
+#ifndef _MSC_VER
+  global_allocation_count++;
+  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Global turn allocation count incremented, now %ld\n", global_allocation_count);
```
eakraly, line 698: This should not be INFO
DEBUG seems like a better fit - would we need a new log line for every allocation? We already have a log line for every allocation

src/apps/relay/userdb.c
```diff
@@ -711,6 +727,14 @@ void release_allocation_quota(uint8_t *user, int oauth, uint8_t *realm) {
     ur_string_map_unlock(rp->status.alloc_counters);
     free(username);
   }
+
+#ifndef _MSC_VER
+  global_allocation_count--;
+  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Global turn allocation count decremented, now %ld\n", global_allocation_count);
```
eakraly, line 733: This can be INFO when the server is in drain mode - I think it will be useful to see in logs that the draining is happening...
But not while in regular operation. We already have too much logging.

src/server/ns_turn_server.c
```diff
@@ -5010,7 +5017,7 @@ void init_turn_server(turn_turnserver *server, turnserver_id id, int verbose, io
 
   server->response_origin_only_with_rfc5780 = response_origin_only_with_rfc5780;
 
-  server->respond_http_unsupported = respond_http_unsupported;
```
eakraly, line 5013: Why change this?

eakraly (commented):
@sgodin a few questions on the PR: code deleted I do not understand and about logging. Thanks!

src/apps/relay/userdb.c
sgodin: I agree, I will make that change.

src/apps/relay/userdb.c
sgodin: Makes sense, I'll make a change for that.

src/server/ns_turn_server.c
sgodin: Nice find - must have been a bad resolution of a merge conflict.  I'll fix it up.

sgodin:
Okay I've made all the changes. However, I can't quite understand why I'm all of a sudden getting clang-format errors on TurnMsgLib.h.  Since I haven't modified this file for my changes, I've even tried using the file checked into master, but it still has issues.  If I make the changes clang-format wants, I ended up with merge conflicts.  Perhaps I need to merge coturn master with my branch again?  Does anyone have any insight?  Probably something simple, that I'm just missing. :)

sgodin:
Ah I see lint / clang-format is failing on master too.  I can't identify a change that would have caused this.  Perhaps the clang-format version was bumped in the CI pipeline??

sgodin:
> @eakraly I have addressed your points with this PR.  Can it be merged in now?  I'm a little baffled at the lint problem (see previous comments), does that need addressing first, or is it a problem that is bigger than this PR?

@eakraly polite ping :)

src/apps/relay/userdb.c
```diff
@@ -711,6 +727,18 @@ void release_allocation_quota(uint8_t *user, int oauth, uint8_t *realm) {
     ur_string_map_unlock(rp->status.alloc_counters);
     free(username);
   }
+
+  int log_level = TURN_LOG_LEVEL_DEBUG;
```
eakraly, line 731: Why changing loglevel? Is it debugging code?

sgodin: @eakraly As a reminder your previous comment was:
"This can be INFO when the server is in drain mode - I think it will be useful to see in logs that the draining is happening...
But not while in regular operation. We already have too much logging."

So I made that change to log at DEBUG normally and at INFO when we are draining.  :)

sgodin: To be clear we are not changing the logging level, we are setting the level at which to send to the TURN_LOG_FUNC.

PR #1535, opened 2024-07-12
Uint16 conversion

WHYHD:
Swapped port vars from int to uint16_t

### DIFF
--- a/src/apps/common/hiredis_libevent2.c
+++ b/src/apps/common/hiredis_libevent2.c
@@ -49,7 +49,7 @@ struct redisLibeventEvents {
   struct event *rev, *wev;
   int rev_set, wev_set;
   char *ip;
-  int port;
+  uint16_t port;
   char *pwd;
   int db;
 };
@@ -213,7 +213,7 @@ void send_message_to_redis(redis_context_handle rch, const char *command, const 
 
 ///////////////////////// Attach /////////////////////////////////
 
-redis_context_handle redisLibeventAttach(struct event_base *base, char *ip0, int port0, char *pwd, int db) {
+redis_context_handle redisLibeventAttach(struct event_base *base, char *ip0, uint16_t port0, char *pwd, int db) {
 
   struct redisLibeventEvents *e = NULL;
   redisAsyncContext *ac = NULL;
@@ -225,7 +225,7 @@ redis_context_handle redisLibeventAttach(struct event_base *base, char *ip0, int
     strncpy(ip, "127.0.0.1", sizeof(ip));
   }
 
-  int port = DEFAULT_REDIS_PORT;
+  uint16_t port = DEFAULT_REDIS_PORT;
   if (port0 > 0) {
     port = port0;
   }
--- a/src/apps/common/ns_turn_utils.c
+++ b/src/apps/common/ns_turn_utils.c
@@ -709,7 +709,7 @@ int get_canonic_origin(const char *o, char *co, int sz) {
               }
             }
 
-            int port = evhttp_uri_get_port(uri);
+            uint16_t port = evhttp_uri_get_port(uri);
             if (port < 1) {
               port = get_default_protocol_port(otmp, schlen);
             }
--- a/src/apps/natdiscovery/natdiscovery.c
+++ b/src/apps/natdiscovery/natdiscovery.c
@@ -55,7 +55,7 @@ static int counter = 0;
 
 #ifdef __cplusplus
 
-static int init_socket(int *socketfd, ioa_addr *local_addr, int local_port, ioa_addr *remote_addr) {
+static int init_socket(int *socketfd, ioa_addr *local_addr, uint16_t local_port, ioa_addr *remote_addr) {
   int ret = 0;
 
   if (local_port >= 0) {
@@ -76,8 +76,8 @@ static int init_socket(int *socketfd, ioa_addr *local_addr, int local_port, ioa_
   return ret;
 }
 
-static int stunclient_send(int sockfd, ioa_addr *local_addr, int *local_port, ioa_addr *remote_addr, int change_ip,
-                           int change_port, int padding, int response_port) {
+static int stunclient_send(int sockfd, ioa_addr *local_addr, uint16_t *local_port, ioa_addr *remote_addr, int change_ip,
+                           uint16_t change_port, int padding, uint16_t response_port) {
   int ret = 0;
 
   turn::StunMsgRequest req(STUN_METHOD_BINDING);
@@ -86,7 +86,7 @@ static int stunclient_send(int sockfd, ioa_addr *local_addr, int *local_port, io
 
   if (response_port >= 0) {
     turn::StunAttrResponsePort rpa;
-    rpa.setResponsePort((uint16_t)response_port);
+    rpa.setResponsePort(response_port);
     try {
       req.addAttr(rpa);
     } catch (turn::WrongStunAttrFormatException &ex1) {
@@ -270,7 +270,7 @@ static int stunclient_receive(int sockfd, ioa_addr *local_addr, ioa_addr *reflex
 }
 
 static int run_stunclient(ioa_addr *local_addr, ioa_addr *remote_addr, ioa_addr *reflexive_addr, ioa_addr *other_addr,
-                          int *local_port, int *rfc5780, int change_ip, int change_port, int padding) {
+                          uint16_t *local_port, int *rfc5780, int change_ip, uint16_t change_port, int padding) {
   int ret = 0;
 
   ret = init_socket(&udp_fd, local_addr, *local_port, remote_addr);
@@ -282,8 +282,8 @@ static int run_stunclient(ioa_addr *local_addr, ioa_addr *remote_addr, ioa_addr 
 }
 
 static int run_stunclient_hairpinning(ioa_addr *local_addr, ioa_addr *remote_addr, ioa_addr *reflexive_addr,
-                                      ioa_addr *other_addr, int *local_port, int *rfc5780, int change_ip,
-                                      int change_port, int padding) {
+                                      ioa_addr *other_addr, uint16_t *local_port, int *rfc5780, int change_ip,
+                                      uint16_t change_port, int padding) {
   int ret = 0;
 
   init_socket(&udp_fd, local_addr, *local_port, remote_addr);
@@ -309,10 +309,10 @@ static int run_stunclient_hairpinning(ioa_addr *local_addr, ioa_addr *remote_add
 }
 
 static int run_stunclient_lifetime(int timer, ioa_addr *local_addr, ioa_addr *remote_addr, ioa_addr *reflexive_addr,
-                                   ioa_addr *other_addr, int *local_port, int *rfc5780, int change_ip, int change_port,
+                                   ioa_addr *other_addr, uint16_t *local_port, int *rfc5780, int change_ip, uint16_t change_port,
                                    int padding) {
   int ret = 0;
-  int response_port;
+  uint16_t response_port;
 
   init_socket(&udp_fd, local_addr, *local_port, remote_addr);
 
@@ -336,7 +336,7 @@ static int run_stunclient_lifetime(int timer, ioa_addr *local_addr, ioa_addr *re
 
 #else
 
-static int init_socket(int *socketfd, ioa_addr *local_addr, int local_port, ioa_addr *remote_addr) {
+static int init_socket(int *socketfd, ioa_addr *local_addr, uint16_t local_port, ioa_addr *remote_addr) {
   int ret = 0;
 
   *socketfd = socket(remote_addr->ss.sa_family, CLIENT_DGRAM_SOCKET_TYPE, CLIENT_DGRAM_SOCKET_PROTOCOL);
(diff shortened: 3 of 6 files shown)


src/apps/natdiscovery/natdiscovery.c
```diff
@@ -55,7 +55,7 @@ static int counter = 0;
 
 #ifdef __cplusplus
 
-static int init_socket(int *socketfd, ioa_addr *local_addr, int local_port, ioa_addr *remote_addr) {
+static int init_socket(int *socketfd, ioa_addr *local_addr, uint16_t local_port, ioa_addr *remote_addr) {
   int ret = 0;
 
   if (local_port >= 0) {
```
jonesmz, line 61: This conditional is now always true.

Make sure you go through the entire data flow for each port variable you are changing to confirm you're properly handling error conditions.

src/apps/natdiscovery/natdiscovery.c
```diff
@@ -86,7 +86,7 @@ static int stunclient_send(int sockfd, ioa_addr *local_addr, int *local_port, io
 
   if (response_port >= 0) {
```
jonesmz, line 87: Another condition that becomes always true

src/apps/relay/dbdrivers/dbd_mysql.c
```diff
@@ -182,9 +182,9 @@ static Myconninfo *MyconninfoParse(char *userdb, char **errmsg) {
       } else if (!strcmp(s, "secret")) {
         co->password = strdup(seq + 1);
       } else if (!strcmp(s, "port")) {
-        co->port = (unsigned int)atoi(seq + 1);
+        co->port = (uint16_t)atoi(seq + 1);
```
jonesmz, line 185: How does string parsing manifest in failure?

Make sure that an invalid entry aborts the operation as soon as possible. Or at least properly communicates to the caller.

src/apps/natdiscovery/natdiscovery.c
WHYHD: You are right. I have reverted the change

src/apps/common/hiredis_libevent2.c
```diff
@@ -49,7 +49,7 @@ struct redisLibeventEvents {
   struct event *rev, *wev;
   int rev_set, wev_set;
   char *ip;
-  int port;
+  uint16_t port;
```
jonesmz, line 52: I think you're going to need to revert this change (revert the whole file).

While an ideal world would have this port be represented as.... a port, which in TCP or UDP protocol networking is always unsigned 16 bits, this variable specifically is used in a third party API (Redis's redisAsyncConnect) which is defined as

```
redisAsyncContext *redisAsyncConnect(const char *ip, int port) {
    redisOptions options = {0};
    REDIS_OPTIONS_SET_TCP(&options, ip, port);
    return redisAsyncConnectWithOptions(&options);
}
```
with `REDIS_OPTIONS_SET_TCP` defined as
```
#define REDIS_OPTIONS_SET_TCP(opts, ip_, port_) do { \
        (opts)->type = REDIS_CONN_TCP;               \
        (opts)->endpoint.tcp.ip = ip_;               \
        (opts)->endpoint.tcp.port = port_;           \
    } while(0)
```
and `redisOptions`  defined as

```

typedef struct {
    /*
     * the type of connection to use. This also indicates which
     * `endpoint` member field to use
     */
    int type;
    /* bit field of REDIS_OPT_xxx */
    int options;
    /* timeout value for connect operation. If NULL, no timeout is used */
    const struct timeval *connect_timeout;
    /* timeout value for commands. If NULL, no timeout is used.  This can be
     * updated at runtime with redisSetTimeout/redisAsyncSetTimeout. */
    const struct timeval *command_timeout;
    union {
        /** use this field for tcp/ip connections */
        struct {
            const char *source_addr;
            const char *ip;
            int port;
        } tcp;
        /** use this field for unix domain sockets */
        const char *unix_socket;
        /**
         * use this field to have hiredis operate an already-open
         * file descriptor */
        redisFD fd;
    } endpoint;

    /* Optional user defined data/destructor */
    void *privdata;
    void (*free_privdata)(void *);

    /* A user defined PUSH message callback */
    redisPushFn *push_cb;
    redisAsyncPushFn *async_push_cb;
} redisOptions;
```


It's *trivially obviously* wrong for this field to use `int`, but that's just how c-language codebases do things in a lot of situations. The extreme amount of manual work needed to do basic type-system style work in c-language is what I would personally point to as the underlying cause for this abuse of the `int` type in this situation, but there's no getting around it.

So for this file, probably better to just leave it as `int` to work with how Redis does it.

Perhaps we can make an improvement here some other time.

src/apps/natdiscovery/natdiscovery.c
jonesmz: To address this you'll need to.

1. Find all places where this function is called, and determine how they could potentially get an invalid port number to pass into this function.
2. It's possible that some callers are *intentionally* passing `-1` into the function to indicate an invalid port. In that case, you'll want to change the parameter to a pointer-to-uint16_t, and use `NULL` to indicate "not passed", instead of `-1`.
3. If none of the callers are intentionally passing invalid ports, but are only able to pass invalid ports because they read an invalid port out of a config file / network packet or something, then you need to put the error handling at the place where the bad valid is acquired.
4. Assuming that none of the callers are intentionally passing invalid ports, then remove the if-check, and make the body of the conditional always happen.

src/apps/natdiscovery/natdiscovery.c
```diff
@@ -336,7 +336,7 @@ static int run_stunclient_lifetime(int timer, ioa_addr *local_addr, ioa_addr *re
 
 #else
 
-static int init_socket(int *socketfd, ioa_addr *local_addr, int local_port, ioa_addr *remote_addr) {
+static int init_socket(int *socketfd, ioa_addr *local_addr, uint16_t local_port, ioa_addr *remote_addr) {
   int ret = 0;
 
   *socketfd = socket(remote_addr->ss.sa_family, CLIENT_DGRAM_SOCKET_TYPE, CLIENT_DGRAM_SOCKET_PROTOCOL);
```
jonesmz, line 342: this function is another place where you have to evaluate all the places where `local_port` is used, and adjust any if-checks that are now always true.

src/apps/natdiscovery/natdiscovery.c
```diff
@@ -356,14 +356,14 @@ static int init_socket(int *socketfd, ioa_addr *local_addr, int local_port, ioa_
...
 
-static int stunclient_send(stun_buffer *buf, int sockfd, ioa_addr *local_addr, int *local_port, ioa_addr *remote_addr,
-                           int change_ip, int change_port, int padding, int response_port) {
+static int stunclient_send(stun_buffer *buf, int sockfd, ioa_addr *local_addr, uint16_t *local_port, ioa_addr *remote_addr,
+                           int change_ip, uint16_t change_port, int padding, uint16_t response_port) {
   int ret = 0;
 
   stun_prepare_binding_request(buf);
 
   if (response_port >= 0) {
```
jonesmz, line 365: this function is another place where you have to evaluate all the places where `local_port` is used, and adjust any if-checks that are now always true.

src/apps/peer/mainudpserver.c
```diff
@@ -54,7 +54,7 @@ static char Usage[] = "Usage: server [options]\n"
 //////////////////////////////////////////////////
 
 int main(int argc, char **argv) {
-  int port = PEER_DEFAULT_PORT;
+  uint16_t port = PEER_DEFAULT_PORT;
```
jonesmz, line 57: on line 80, this variable is replaced with the result of calling `atoi` which returns a signed `int`. That could result in a negative int, or an int that is above the range of uint16_t, implicitly converting into a uint16_t, which would silently lead to the wrong behavior.

The call to `atoi` needs to either:

1. be replaced with a more appropriate function which indicates an error if the provided string is not parsable as an integer, with additional error checking regarding the value not exceeding what a uint16_t can hold
2. rework how `atoi` is used to do appropriate error checking of the result.

jonesmz:
in many places in this review `port` is passed to `make_ioa_addr` which is defined in https://github.com/coturn/coturn/blob/master/src/client/ns_turn_ioaddr.c#L214, but you're not changing that function to take a uint16_t. That's one of the most "core" functions dealing with ports in the coturn code, so it's probably the better place to start.

You can first address that function, and then sweep "out" from there, adjusting error handling at each place.

You can see at https://learn.microsoft.com/en-us/windows/win32/api/ws2def/ns-ws2def-sockaddr_in and https://man7.org/linux/man-pages/man7/ip.7.html (with in_port_t defined here as uint16_t https://pubs.opengroup.org/onlinepubs/009695399/basedefs/netinet/in.h.html ) that the actual socket api natively works with unsigned 16bits.

src/apps/relay/dbdrivers/dbd_redis.c
```diff
@@ -134,9 +134,9 @@ static Ryconninfo *RyconninfoParse(const char *userdb, char **errmsg) {
       } else if (!strcmp(s, "secret")) {
         co->password = strdup(seq + 1);
       } else if (!strcmp(s, "port")) {
-        co->port = (unsigned int)atoi(seq + 1);
```
jonesmz, line 137: I recommend caution in changing the dbdrivers code. It's not guaranteed that a database protocol uses the same underlying concept of a port that the TCP and UDP protocols use.

jonesmz: You might want to defer any changes to the dbdrivers to another pull request in the future.

src/apps/relay/dbdrivers/dbd_mysql.c
```diff
@@ -44,7 +44,7 @@ struct _Myconninfo {
   char *dbname;
   char *user;
   char *password;
-  unsigned int port;
+  uint16_t port;
```
jonesmz, line 47: I recommend caution in changing the dbdrivers code. It's not guaranteed that a database protocol uses the same underlying concept of a port that the TCP and UDP protocols use.


You might want to defer any changes to the dbdrivers to another pull request in the future.

src/apps/common/ns_turn_utils.c
```diff
@@ -709,7 +709,7 @@ int get_canonic_origin(const char *o, char *co, int sz) {
               }
             }
 
-            int port = evhttp_uri_get_port(uri);
+            uint16_t port = evhttp_uri_get_port(uri);
```
jonesmz, line 712: does `evhttp_uri_get_port` return a uint16_t? i don't see it changed in this PR.

src/apps/common/ns_turn_utils.c
```diff
@@ -709,7 +709,7 @@ int get_canonic_origin(const char *o, char *co, int sz) {
...
             if (port < 1) {
```
jonesmz, line 713: Now that `port` is uint16_t, this would be more meaningfully expressed as `port != 0`

src/apps/common/ns_turn_utils.c
```diff
@@ -709,7 +709,7 @@ int get_canonic_origin(const char *o, char *co, int sz) {
...
               port = get_default_protocol_port(otmp, schlen);
```
jonesmz, line 714: does `get_default_protocol_port` return uint16_t? i don't see it changed in this PR.

src/apps/peer/udpserver.c
```diff
@@ -112,8 +112,12 @@ static int udp_create_server_socket(server_type *server, const char *ifname, con
   return 0;
 }
 
-static server_type *init_server(int verbose, const char *ifname, char **local_addresses, size_t las, int port) {
+static server_type *init_server(int verbose, const char *ifname, char **local_addresses, size_t las, uint16_t port) {
 
+  if(port == 65535){
```
jonesmz, line 117: i think you probably want `USHRT_MAX` instead of 65535.

or `0xffff`

src/apps/peer/udpserver.c
```diff
@@ -112,8 +112,12 @@ static int udp_create_server_socket(server_type *server, const char *ifname, con
...
+    return NULL;
+  }
+  
   server_type *server = (server_type *)malloc(sizeof(server_type));
 
   if (!server) {
```
jonesmz, line 123: instead of `return server` here, it should probably be `return NULL` to more clearly indicate we're returning due to allocation failure.

src/apps/peer/udpserver.c
```diff
@@ -112,8 +112,12 @@ static int udp_create_server_socket(server_type *server, const char *ifname, con
   return 0;
 }
 
-static server_type *init_server(int verbose, const char *ifname, char **local_addresses, size_t las, int port) {
+static server_type *init_server(int verbose, const char *ifname, char **local_addresses, size_t las, uint16_t port) {
```
jonesmz, line 116: add a comment indicating why port being USHRT_MAX means we should return early.

ggarber:
This PR has been opened for 1y without response tothe comments from the reviewer.  If it is important please address those comments or at least reply to them or i´ll close this PR  next week.    Thank you very much for your work on this!

eakraly:
will be done in #1752

PR #1540, opened 2024-07-22
Fix #1528

### LINKED ISSUE #1528: Specifying --log-file=stdout still generates a log file in the current execution path.
system: Windows 10
coturn: 4.6.2
config:
`
# Coturn TURN SERVER configuration file
listening-port=8892
listening-ip=0.0.0.0
external-ip=x.x.x.x
verbose
lt-cred-mech
#use-auth-secret
user=user:pwd
realm=WDP5
no-tls
no-dtls
no-cli
# fingerprint
syslog
no-stdout-log
log-file=stdout
`

image

How to set up the log not to generate a log file but only output to stdout?

KangLin:


src/apps/common/apputils.c
```diff
@@ -898,7 +898,7 @@ int getdomainname(char *name, size_t len) {
 
   dw = DsRoleGetPrimaryDomainInformation(NULL, DsRolePrimaryDomainInfoBasic, (PBYTE *)&info);
   if (dw != ERROR_SUCCESS) {
-    TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "DsRoleGetPrimaryDomainInformation: %u\n", dw);
+    //TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "DsRoleGetPrimaryDomainInformation: %u\n", dw);
```
jonesmz, line 901: I don't think commenting the log statements out is the right approach.

PR #1548, opened 2024-07-29
Fix nodejs/glibc problem with old container images.

jonesmz:
This problem is caused by this issue: https://github.com/actions/checkout/issues/1809

Several comments include documentation on various environment variables to force it to use the older nodejs release still, but probably those various workarounds will stop working eventually.


jonesmz:
an alternative to https://github.com/coturn/coturn/pull/1543

eakraly (approved):

eakraly:
Thanks @jonesmz !

PR #1553, opened 2024-08-04
Use bool, instead of int, for the functions in ns_turn_msg.c

jonesmz:
And address knockon effects in other files, e.g. adjust if-statements and other function parameters and return types.

### DIFF
--- a/src/apps/common/stun_buffer.c
+++ b/src/apps/common/stun_buffer.c
@@ -66,36 +66,36 @@ void stun_tid_generate_in_message(stun_buffer *buf, stun_tid *id) {
 
 ////////////////////////////////////////////////////////
 
-static inline int is_channel_msg(const stun_buffer *buf) {
+static inline bool is_channel_msg(const stun_buffer *buf) {
   if (buf && buf->len > 0) {
     return is_channel_msg_str(buf->buf, (size_t)(buf->len));
   }
-  return 0;
+  return false;
 }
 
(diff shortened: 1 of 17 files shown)


eakraly:
Hi @jonesmz !
I went over the PR as much as I could but there are so many changes...
Not that I have anything to comment on but how confident are you with this change?

jonesmz:
As reasonably confident as I can be.

I agree its a lot of individual changes, but given the nature of c-language type conversions, the risk isn't ezclusively or even mostly in what was changed, but in the possibility of code that wasnt changed but should have been.

My methodology was:

For each function in the c-file:

1. Adjust the function to return bool/accept book params
2. use my ide's search function to locate every string match for the functions name
3. Adjust every located use of the function.
4. For functions who's return type is impacted by that adjustment, goto 2.

The vast majority of code outside of ns_turn_msg that was impacted was if statements that were checking if the returned value was less than 0 or more than 0. I think a big part of the risk there is that these weren't consistent. Sometimes 0 meant success, other times it meant failure. So while I'm pretty sure I kept it consistent, I could have reversed a conditional by mistake and we wouldn't ever get a warning about it.

The other big risk is that our unit testing capabilities are pretty low in general. I have some unit tests at work that are for my own implentatipn of stun message parsing, but I'd have to adjust the code a lot (entirely different api, c++ instead of c) and need to get permission from my legal team to share them.

eakraly (approved):

PR #1555, opened 2024-08-04
Fix compiler warnings from continuous integration

jonesmz:
Almost all of the warnings were about truncating pointers, because sizeof(void*) != sizeof(long) on all platforms.


eakraly (approved):
Nice!

Issue #1475, opened 2024-04-23
Serve prometheus on localhost


alexg-axis:
The `promhttp` library is hard coded to serve the metrics on all interfaces. As such the metrics are open to all by default, with no way of being controlled. The underlying `microhttpd` supports several options, though. The prom library is however unmaintained and putting effort there to add additional options would likely not be a worth investment. I therefore suggest to implement the HTTP part of the library (basically a call to start the HTTP daemon and serve the `/metrics` endpoint) to coturn instead, so that we're free to add additional options that better suite the coturn use case. I'm willing to put in the required work.

PR #1591, opened 2024-10-24
Implement custom prometheus http handler

alexg-axis:
Implement a custom prometheus http handler in order to:

1. Support listening on a specified address as opposed to any
2. Remove the requirement on the unmaintained promhttp library

This feature comes with one limitation: if an IPv4 address is used, the server will not listen on the IPv6-mapped address, even if IPv6 is available. That is, dual-stacking does not work.

Solves: #1475

### DIFF
--- a/examples/run_tests_prom.sh
+++ b/examples/run_tests_prom.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+function assert_prom_no_response() {
+  wget --quiet --output-document=/dev/null --tries=1 "$1"
+  status="$?"
+  if [ "$status" -eq 0 ]; then
+    echo FAIL
+    exit 1
+  else
+    echo OK
+  fi
+}
+
+function assert_prom_response() {
+  # Match something that looks like the expected body
+  wget --quiet --output-document=- --tries=1 "$1" | grep 'TYPE\|HELP\|counter\|gauge' >/dev/null
+  status="$?"
+  if [ "$status" -eq 0 ]; then
+    echo OK
+  else
+    echo FAIL
+    exit "$status"
+  fi
+}
+
+echo "Running without prometheus"
+../bin/turnserver &
+turnserver_pid="$!"
+sleep 2
+assert_prom_no_response "http://localhost:9641/metrics"
+kill "$turnserver_pid"
+
+echo "Running turnserver with prometheus, using defaults"
+../bin/turnserver --prometheus &
+turnserver_pid="$!"
+sleep 2
+assert_prom_response "http://localhost:9641/metrics"
+kill "$turnserver_pid"
+
+echo "Running turnserver with prometheus, using custom address"
+../bin/turnserver --prometheus --prometheus-address="127.0.0.1" &
+turnserver_pid="$!"
+sleep 2
+assert_prom_response "http://127.0.0.1:9641/metrics"
+kill "$turnserver_pid"
+
+echo "Running turnserver with prometheus, using custom port"
+../bin/turnserver --prometheus --prometheus-port="8080" &
+turnserver_pid="$!"
+sleep 2
+assert_prom_response "http://localhost:8080/metrics"
+kill "$turnserver_pid"
+
+echo "Running turnserver with prometheus, using custom address and port"
+../bin/turnserver --prometheus --prometheus-address="127.0.0.1" --prometheus-port="8080" &
+turnserver_pid="$!"
+sleep 2
+assert_prom_response "http://127.0.0.1:8080/metrics"
+kill "$turnserver_pid"
--- a/src/apps/relay/prom_server.c
+++ b/src/apps/relay/prom_server.c
@@ -35,6 +35,42 @@ prom_counter_t *turn_total_traffic_peer_sentb;
 
 prom_gauge_t *turn_total_allocations;
 
+#if MHD_VERSION >= 0x00097002
+#define MHD_RESULT enum MHD_Result
+#else
+#define MHD_RESULT int
+#endif
+
+MHD_RESULT promhttp_handler(void *cls, struct MHD_Connection *connection, const char *url, const char *method,
+                            const char *version, const char *upload_data, size_t *upload_data_size, void **con_cls) {
+  MHD_RESULT ret;
+
+  char *body = "not found";
+  enum MHD_ResponseMemoryMode mode = MHD_RESPMEM_PERSISTENT;
+  unsigned int status = MHD_HTTP_NOT_FOUND;
+
+  if (strcmp(method, "GET") != 0) {
+    status = MHD_HTTP_METHOD_NOT_ALLOWED;
+    body = "method not allowed";
+  } else if (strcmp(url, "/metrics") == 0) {
+    body = prom_collector_registry_bridge(PROM_COLLECTOR_REGISTRY_DEFAULT);
+    mode = MHD_RESPMEM_MUST_FREE;
+    status = MHD_HTTP_OK;
+  }
+
+  struct MHD_Response *response = MHD_create_response_from_buffer(strlen(body), body, mode);
+  if (response == NULL) {
+    if (mode == MHD_RESPMEM_MUST_FREE) {
+      free(body);
+    }
+    ret = MHD_NO;
+  } else {
+    ret = MHD_queue_response(connection, status, response);
+    MHD_destroy_response(response);
+  }
+  return ret;
+}
+
 void start_prometheus_server(void) {
   if (turn_params.prometheus == 0) {
     TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "prometheus collector disabled, not started\n");
@@ -103,13 +139,8 @@ void start_prometheus_server(void) {
   turn_total_allocations = prom_collector_registry_must_register_metric(
       prom_gauge_new("turn_total_allocations", "Represents current allocations number", 1, typeLabel));
 
-  promhttp_set_active_collector_registry(NULL);
-
   // some flags appeared first in microhttpd v0.9.53
   unsigned int flags = 0;
-  if (MHD_is_feature_supported(MHD_FEATURE_IPv6) && is_ipv6_enabled()) {
-    flags |= MHD_USE_DUAL_STACK;
-  }
 #if MHD_VERSION >= 0x00095300
   flags |= MHD_USE_ERROR_LOG;
 #endif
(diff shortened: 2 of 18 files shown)


eakraly:
Thanks @alexg-axis  - this is an awesome contribution!
Getting rid of libpromhttp gets us half way to remove unsupported libraries from DO

src/apps/relay/prom_server.c
```diff
@@ -127,7 +161,36 @@ void start_prometheus_server(void) {
     TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING, "prometheus exporter server will start using SELECT. "
                                           "The exporter might be unreachable on highly used servers\n");
   }
-  struct MHD_Daemon *daemon = promhttp_start_daemon(flags, turn_params.prometheus_port, NULL, NULL);
+
+  ioa_addr server_addr;
+  addr_set_any(&server_addr);
+  if (turn_params.prometheus_address[0]) {
```
eakraly, line 167: What about default? Maybe having it open to localhost?

src/apps/relay/prom_server.c
```diff
@@ -1,5 +1,6 @@
 #include "prom_server.h"
 #include "mainrelay.h"
+#include "microhttpd.h"
```
eakraly, line 3: Already included in .h file - why another include?

eakraly:
Also, could you please add some test that validates prometheus interface is up and running?  See examples folder where you can add another test
Also, you missed a reference to promhttp in .github folder

src/apps/relay/prom_server.c
alexg-axis: Good spot. Removed.

src/apps/relay/prom_server.c
```diff
@@ -35,6 +35,44 @@ prom_counter_t *turn_total_traffic_peer_sentb;
 
 prom_gauge_t *turn_total_allocations;
 
+#if MHD_VERSION >= 0x00097500
```
eakraly, line 38: Seems like MHD_Result was introduced in libmicrohttpd 4 years ago in 0.9.71, not 0.9.75
https://github.com/Karlson2k/libmicrohttpd/commit/6347f514aa2388e774d5bf356df8046864e5f73c

Also, this was not in promhttp - so we are not losing anything by not having `#if MHD_VERSION >= 0x00097500`

I suggest to simplify the code and require libmicrohttpd >= 0.9.71
What do you think?

alexg-axis: Yeah, I think it makes a lot of sense. There's some other code also checking the version of microhttpd, like the error log and epoll flags. Should we remove those ifdefs in this PR or take that another time?

eakraly: Please remove it now - no point in adding and then removing code.
Thanks!

alexg-axis: The CI is installing libmicrohttpd 0.9.66-1, seemingly due to ubuntu:20.04 being used. If I try to run ubuntu:22.04 instead, I get 0.9.75. 

I think I've seen a few issues and PRs on this topic. I guess the "easiest" thing to do is to update the CI/CD and start requiring newer versions of ubuntu. Otherwise we should just be able to revert my latest commit, again adding support for <0.9.71.

eakraly: Ugh. Let's use `int` return type instead of `MHD_Status` - ugly but should work I hope.

eakraly: I see. Let's do that (revert). Maybe you can make it a little nicer by defining `MHD_Result` as int for older version and keeping the code in its new form
```
#if MHD_VERSION < 0x00097500
define MHD_Result int
#endif

MHD_Result promhttp_handler(....
```

alexg-axis: Note that it's `enum MHD_Result`. I defined a `MHD_RESULT` that works the way you described. Does that work for you?

Also, I stumbled upon a changelog entry for the change. It recommends using `MHD_VERSION >= 0x00097002`. So I changed that as well. Here's a link to the changelog: https://github.com/Karlson2k/libmicrohttpd/blob/a1085032dc631e387eb8e06b41a382e7dc6dd8eb/ChangeLog#L1055.

alexg-axis:
@eakraly I noticed a potentially problematic behavior I can't really explain. I don't think it's related to these changes, but it's worth bringing up.

Every now and then the server seems to return empty metrics.

For example:

```
curl --silent localhost:9641/metrics | grep -A1 turn_total_traffic_rcvb
# HELP turn_total_traffic_rcvb Represents total finished sessions received bytes
# TYPE turn_total_traffic_rcvb counter
turn_total_traffic_rcvb 1484
```

```
curl --silent localhost:9641/metrics | grep -A1 turn_total_traffic_rcvb
# HELP turn_total_traffic_rcvb Represents total finished sessions received bytes
# TYPE turn_total_traffic_rcvb counter
turn_total_traffic_rcvb 1484
```

```
curl --silent localhost:9641/metrics | grep -A1 turn_total_traffic_rcvb
# HELP turn_total_traffic_rcvb Represents total finished sessions received bytes
# TYPE turn_total_traffic_rcvb counter
```

```
curl --silent localhost:9641/metrics | grep -A1 turn_total_traffic_rcvb
# HELP turn_total_traffic_rcvb Represents total finished sessions received bytes
# TYPE turn_total_traffic_rcvb counter
turn_total_traffic_rcvb 1484
```

When the prom lib is generating the response body, there are a few early returns that I guess could result in metrics missing. The error code is unhandled by `prom_collector_registry_bridge`.

https://github.com/digitalocean/prometheus-client-c/blob/c57034d196582d99267d027abb52a05a55dc07f6/prom/src/prom_collector_registry.c#L215

In grafana, for example, I guess it's quite common to fill in "null" values by repeating the last value. But perhaps this is a symptom of something else?

eakraly:
> @eakraly I noticed a potentially problematic behavior I can't really explain. I don't think it's related to these changes, but it's worth bringing up. I've only noticed it once, so it might just be due to something unique to my environment.
> 
> Every now and then the server seems to return empty metrics.

Interesting. Do you observe it without your change? I bet it is unrelated. If it is - we can go on and create another issue to tackle this specific problem.

alexg-axis:
@eakraly I've now built and tested the code manually with asan and things work as expected without any issues. I have also added an option to control the base path (or the full path). This helps the prometheus server to run behind a reverse proxy on a specified path as coturn can be told to use that path.

In https://github.com/coturn/coturn/pull/1591/commits/e5cec04403539ce513b0ed37bf8e0724480fdedb I've opted for specifying a base path and "hard coding" the `/metrics` path. But as that is the only path today, I also wrote https://github.com/coturn/coturn/pull/1591/commits/1535c9605b473a2ad429b5085c4e4531704ce89d to allow users to specify the full path instead.

What are your thoughts on that option? I want to be able to expose prometheus behind a reverse proxy on a specified path and it would be nice if that path could for example be `/metrics/coturn` without coturn requiring `/metrics` to be added to the end, ending up with the repetitive `/metrics/coturn/metrics` path.

One potential issue I see is that it might be more cumbersome to add additional endpoints to the server handler (whatever they may be) as now `/metrics` is a common base path for all endpoints, so adding an endpoint would result in `/metrics/xyz`, where you might have wanted `/metrics` and `/xyz`.

eakraly:
Thanks @alexg-axis !

1) Having default to be equal to today's configuration is great - that means no breaking change.
2) I've seen a lot of software that exposes metrics on `/metrics` because that's the prometheus default and when you scrape multiple targets - they have to expose on different ports anyway. Also, it means that prometheus need to scrape multiple targets (let's say coturn and node_exporter) anyway - whether they expose it on `/metric/coturn` or `/metrics`.

> What are your thoughts on that option?

I'm ok with the change - IFF that solves a problem for you (though I do not really see what it would be - if you can expand on your reverse proxy config and how they will work together better it would be great)

> One potential issue I see is ....

I do not understand your concern - the handler you edited is prom specific HTTP handler. In any case, this is a simple if..else flow that can match more paths in the future (let's say we split prom into 2 subsets of metrics /metrics/a and /metrics/b - we can still match them).

eakraly:
@alexg-axis my additional suggestion would be to split this PR into
1) Custom prom http handler (as it was before last 2 commits)
2) Custom prom path setting (last 2 commits)

These are 2 different things and better have them separately (just in case there is a fallout and we need to bisect the regression)

eakraly:
> > @alexg-axis my additional suggestion would be to split this PR into
> > 
> > 1. Custom prom http handler (as it was before last 2 commits)
> > 2. Custom prom path setting (last 2 commits)
> > 
> > These are 2 different things and better have them separately (just in case there is a fallout and we need to bisect the regression)
> 
> I've rebased this PR and removed the addition of base path handling. Let's take that discussion separately as you said ( #1602. I'll rebase once this is in). Regarding testing, could you clarify what you would want to add before we can get this in?
> 
> Would you like me to change `examples/run_tests.sh` to start turnserver with prometheus enabled on localhost for example? Or would you like an additional `examples/run_tests_prom.sh` that starts prometheus with the different ways to configure prometheus?

For testing I think separate `examples/run_tests_prom.sh` would be better - prom is not built for all and we may see a bunch of failures changing existing test setup. Creating new one gives us opportunity to expand for more prom tests in the future.

alexg-axis:
> > > @alexg-axis my additional suggestion would be to split this PR into
> > > 
> > > 1. Custom prom http handler (as it was before last 2 commits)
> > > 2. Custom prom path setting (last 2 commits)
> > > 
> > > These are 2 different things and better have them separately (just in case there is a fallout and we need to bisect the regression)
> > 
> > 
> > I've rebased this PR and removed the addition of base path handling. Let's take that discussion separately as you said ( #1602. I'll rebase once this is in). Regarding testing, could you clarify what you would want to add before we can get this in?
> > Would you like me to change `examples/run_tests.sh` to start turnserver with prometheus enabled on localhost for example? Or would you like an additional `examples/run_tests_prom.sh` that starts prometheus with the different ways to configure prometheus?
> 
> For testing I think separate `examples/run_tests_prom.sh` would be better - prom is not built for all and we may see a bunch of failures changing existing test setup. Creating new one gives us opportunity to expand for more prom tests in the future.

I've added a few tests. Could you trigger the CI pipeline so we can see them work there as well?

examples/run_tests_prom.sh
```diff
@@ -0,0 +1,59 @@
...
+    echo FAIL
+    exit 1
+  else
+    echo OK
+  fi
+}
+
+function assert_prom_response() {
+  # Match something that looks like the expected body
+  curl --silent --fail "$1" | grep 'TYPE\|HELP\|counter\|gauge' >/dev/null
```
eakraly, line 16: Please use wget instead of curl - it is not installed

alexg-axis: @eakraly is there anything else needed from our side?

examples/run_tests_prom.sh
```diff
@@ -0,0 +1,59 @@
...
+  if [ "$status" -eq 0 ]; then
+    echo OK
+  else
+    echo FAIL
+    exit "$status"
+  fi
+}
+
+echo "Running without prometheus"
+../bin/turnserver &
```
eakraly, line 27: > /dev/null

eakraly, line 27: > /dev/null

examples/run_tests_prom.sh
```diff
@@ -0,0 +1,59 @@
...
+turnserver_pid="$!"
+sleep 2
+assert_prom_no_response "http://localhost:9641/metrics"
+kill "$turnserver_pid"
+
+echo "Running turnserver with prometheus, using defaults"
+../bin/turnserver --prometheus &
```
eakraly, line 34: > /dev/null

examples/run_tests_prom.sh
```diff
@@ -0,0 +1,59 @@
...
+turnserver_pid="$!"
+sleep 2
+assert_prom_response "http://localhost:9641/metrics"
+kill "$turnserver_pid"
+
+echo "Running turnserver with prometheus, using custom address"
+../bin/turnserver --prometheus --prometheus-address="127.0.0.1" &
```
eakraly, line 41: > /dev/null

examples/run_tests_prom.sh
```diff
@@ -0,0 +1,59 @@
...
+turnserver_pid="$!"
+sleep 2
+assert_prom_response "http://127.0.0.1:9641/metrics"
+kill "$turnserver_pid"
+
+echo "Running turnserver with prometheus, using custom port"
+../bin/turnserver --prometheus --prometheus-port="8080" &
```
eakraly, line 48: > /dev/null

eakraly (approved):

alexg-axis:
@eakraly thanks for merging! I've updated https://github.com/coturn/coturn/pull/1602 to add a flag to specify at which path the metrics should be served.

PR #1598, opened 2024-11-18
Reverse SOFTWARE_ATTRIBUTE_OPT to avoid inverse logic

eakraly:
As part of looking at # , I figured that sending SOFTWARE_ATTRIBUTE is also part of a problem as it increases messages sent out by coturn and thus increasing amplification factor.
With SOFTWARE_ATTRIBUTE:
WithOUT SOFTWARE_ATTRIBUTE:

As I observed with multiple providers using coturn - some of the are sending it. Meaning, they do not set `--no-software-attribute` - most probably due to lack of clarity about this setting.

I believe sending SOFTWARE_ATTRIBUTE should be off by default - which is hinted in the RFC (https://datatracker.ietf.org/doc/html/rfc8489#section-16.1.2)

More changes:
- Extract setting the attribute into a function to avoid code duplication
- This option is now not reloadable
- The option is now called `software_attribute` because inverse logic creates multiple double-not in the code which makes it harder to read.
- `no-software_attribute` is still functional but marked as deprecated in documentation

Test Plan:
- Run local tests with different cli arguments (new and deprecated) and confirm SOFTWARE attribute is off by default, and added when arguments say so


eakraly:
This is a breaking change so delaying merging for a little

src/server/ns_turn_server.c
```diff
@@ -75,13 +75,23 @@ static inline int get_family(int stun_family, ioa_engine_handle e, ioa_socket_ha
 ////////////////////////////////////////////////
 
 const char *get_version(turn_turnserver *server) {
-  if (server && !*server->no_software_attribute) {
+  if (server && server->software_attribute) {
```
jonesmz, line 78: you're dropping a dereference on this line.

Was the dereference a bug?

jonesmz (approved):

PR #1601, opened 2024-11-21
Add MariaDB to README.md

robertsilen:
Suggesting to add MariaDB to README - any reason not to?


eakraly:
Thanks @robertsilen 
It does sound like no reason not to have it

PR #1605, opened 2024-11-26
Alternate server tcp udp distinction

### LINKED ISSUE #1604: Alternate Server Bug: Separate Counters for TCP, UDP, and TLS to Prevent Misdirected Connections
The 300 Try Alternate response from the Coturn server includes Alternate Server data if configured. When using the default port (3478) or when the same port is configured for both TCP and UDP, everything works fine.

However, a problem arises when different ports are used for UDP and TCP. The Alternate Server mechanism cannot distinguish between UDP and TCP ports. For example, suppose we configure listening ports as 443 for TCP and 3478 for UDP. In this case, the Coturn server sending the 300 Try Alternate response (the redirector server) cannot determine whether an alternate server address in the list corresponds to UDP or TCP. As a result, when a client reaches this redirector server, it might receive a response pointing to a TCP or UDP server at random. If a client using TCP tries to connect to the UDP port of the second Coturn server (the main worker), the communication will fail, and vice versa.

Port 443 for TCP is one of the most commonly used ports in corporate network environments. However, UDP is often blocked, especially port 443 UDP, as it is a low-numbered port and subject to restrictions. Therefore, for most scenarios, it is necessary to separate TCP on port 443 and UDP on port 3478.

To address this issue, I prepared an auxiliary mechanism. It is straightforward and does not affect existing mechanisms such as Alternate Server, Aux Server, or TLS Alternate Server. I simply added separate TCP and UDP alternate server lists without altering the old mechanisms.

There was just one exception: a bug in the TLS Alternate Server mechanism as far as I can see. Previously, the round-robin selection algorithm only used a single counter shared by the Alternate Server list, even though there was already a dedicated TLS counter available but unused. I modified the mechanism to utilize the existing TLS counter for TLS Alternate Servers, ensuring proper functionality. Similarly, I ensured separate counters are now used for TCP and UDP lists.

Cybermilitia:
- Added separate TCP and UDP alternate server lists to resolve port ambiguity.
- Enabled the use of an existing TLS counter for Alternate Server round-robin selection, fixing a bug where only the generic counter was used.
- Ensures correct redirection for clients using different protocols (TCP/UDP).
- Improves reliability and prevents misdirected connections in multi-protocol setups.

Fixes #1604


Cybermilitia:
As far as I can understand, the fails are not related to the code directly. Do you have any opinion?

eakraly:
Hi @Cybermilitia ! Could you please rebase and resolve the conflicts? Thanks!

ggarber:
Fixed bug with ports and also rebased and reformatted.   But it is still missing updating the config file and the READMe doc.

PR #1609, opened 2024-12-11
Update version to 4.6.3

eakraly:
preparing for new tag 4.6.3

### DIFF
--- a/docker/coturn/Makefile
+++ b/docker/coturn/Makefile
@@ -24,7 +24,7 @@ dockerify = $(strip $(if $(call eq,$(1),linux/arm32v6),linux/arm/v6,\
 # Project parameters #
 ######################
 
-COTURN_VER ?= 4.6.3
+COTURN_VER ?= 4.6.2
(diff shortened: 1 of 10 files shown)


docker/coturn/Makefile
```diff
@@ -24,7 +24,7 @@ dockerify = $(strip $(if $(call eq,$(1),linux/arm32v6),linux/arm/v6,\
 # Project parameters #
 ######################
 
-COTURN_VER ?= 4.6.2
+COTURN_VER ?= 4.6.3
```
tyranron: For Docker image there is no point to change the version, since it's versioned a little bit different from the sources. It will be changed in a separate commit preparing the image release.

tyranron (approved):
@eakraly I've removed from changelog Docker-related stuff (and some unrelevant for end users CI stuff). Everything else seems LGTM.

docker/coturn/Makefile
eakraly: Sure - just looks better when searching up the code and then seeing old version number is weird

Issue #1629, opened 2025-01-16
Segmentation fault on SIGHUP


nicolas-laduguie:
Hi, 

Our Coturn V4.6.3 (latest at the time of writing) compiled from source receives segfault when triggering SIGHUP signal on the process.
From the logs it shows:

```
Jan 16 09:42:53 XXX
 kernel: [7254972.859762] turnserver[3777793]: segfault at 7ffff4e00c98 ip 00005555555a9d1e sp 00007ffff4e00ca0 error 6 in turnserver[55555555d000+6e000] likely on CPU 0 (core 0, socket 0)
Jan 16 09:42:53 XXX kernel: [7254972.859786] Code: 4c 8d 05 ad e9 02 00 48 8d 05 b6 e7 02 00 48 89 c1 ba 00 00 00 00 be 71 01 00 00 48 8d 05 02 e7 02 00 48 89 c7 b8 00 00 00 00 <e8> 56 09 00 00 e8 00 fc ff ff c7 05 b2 1c 07 00 00 00 00 00 8b 05
```

Could capture what happens with GDB and here is what I get:
```
Thread 4 "turnserver" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffe7600640 (LWP 3778663)]
0x00005555555a9d1e in set_rtpfile () at src/apps/common/ns_turn_utils.c:369
369	    TURN_LOG_FUNC(TURN_LOG_LEVEL_DEBUG, "%s: resetting the log file\n", __FUNCTION__);
```

Seems that resetting the log file doesn't work properly, but don't understand why.
Anyone able to help me ?

System information:
6.8.0-47-generic #47~22.04.1-Ubuntu
AMD64

nicolas-laduguie:
Backtrace capture:

```
(gdb) bt
#0  0x00005555555a9d1e in set_rtpfile () at src/apps/common/ns_turn_utils.c:369
#1  0x00005555555aaa92 in turn_log_func_default (file=0x5555555d8418 "src/apps/common/ns_turn_utils.c", line=369, level=TURN_LOG_LEVEL_DEBUG, format=0x5555555d84b8 "%s: resetting the log file\n") at src/apps/common/ns_turn_utils.c:607
#2  0x00005555555a9d23 in set_rtpfile () at src/apps/common/ns_turn_utils.c:369
[...]

PR #1630, opened 2025-01-16
Fix infinite loop while triggering SIGHUP and log file reset causing …

nicolas-laduguie:
…segmentation fault

This reverts commit 8f9446cf8fe4b48ae542c630416fac63d06a6802.


nicolas-laduguie:
Should fix #1629

jonesmz:
Is it also possible for you to add a unit test for this scenario?

nicolas-laduguie:
@jonesmz unfortunately I am not familiar enough with the project and the testing mechanism in place.
Furthermore, that would be an integration test required to assess the process is not crashing.

Issue #1628, opened 2025-01-13
CPUs number as config var


shavnevnikita:
Issue: in virtualized env (cgroups) `_SC_NPROCESSORS_CONF` and `_SC_NPROCESSORS_ONLN` may contain not adequate values.

PR #1640, opened 2025-01-30
CPUs number as config var

shavnevnikita:
Implemented feature mentioned in #1628


src/apps/relay/mainrelay.c
```diff
@@ -2358,6 +2362,16 @@ static void set_option(int c, char *value) {
   case RESPOND_HTTP_UNSUPPORTED_OPT:
     turn_params.respond_http_unsupported = get_bool_value(value);
     break;
+  case CPUS_NUMBER_OPT:
+    if (atoi(value) > MAX_CPUS_NUMBER) {
+      TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING, "max number of cpus is 128.\n");
```
uninvited, line 2367: `TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING, "Max number of cpus is %d", MAX_CPUS_NUMBER);`
?

src/apps/relay/mainrelay.c
```diff
@@ -162,9 +160,9 @@ turn_params_t turn_params = {
     NULL, /*relay_addrs*/
     0,    /*default_relays*/
 
-    NULL,                                 /*external_ip*/
-    DEFAULT_GENERAL_RELAY_SERVERS_NUMBER, /*general_relay_servers_number*/
-    0,                                    /*udp_relay_servers_number*/
+    NULL, /*external_ip*/
+    255,  /*general_relay_servers_number*/
```
eakraly, line 164: Why change this code?

src/apps/relay/mainrelay.c
```diff
@@ -3045,6 +3042,26 @@ int main(int argc, char **argv) {
...
+    } else if (turn_params.cpus > MAX_CPUS_NUMBER) {
+      turn_params.cpus = MAX_CPUS_NUMBER;
+    }
+  }
+
+  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "System cpu num is %lu\n", get_system_number_of_cpus());
+  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "System enable num is %lu\n", get_system_active_number_of_cpus());
+  TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "Will be used %lu\n", turn_params.cpus);
+
+  if (turn_params.general_relay_servers_number == 255) {
```
eakraly, line 3061: This change does nothing - if will always be executed as before

src/apps/relay/mainrelay.c
shavnevnikita: Cause we need to distinguish between explicitly set relay-threads to 0 and not set.

src/apps/relay/mainrelay.c
shavnevnikita: Now we are checking general_relay_servers_number after first pass of reading config, so if it is set we dont wanna override it.

PR #1653, opened 2025-03-14
Add Remote IP on log file for invalid user or connexion

drmanu011:
Added Remote IP logging on TURN_LOG_FUNC for invalid user, needed for protect turnserver with fail2ban

Are you OK?

### DIFF
--- a/src/server/ns_turn_server.c
+++ b/src/server/ns_turn_server.c
@@ -3400,8 +3400,10 @@ static int check_stun_auth(turn_turnserver *server, ts_ur_super_session *ss, stu
   usname[alen] = 0;
 
   if (!is_secure_string(usname, 1)) {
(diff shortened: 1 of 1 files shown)


eakraly:
Hi @drmanu011 

2 comments:
1. It will probably not help as you may block huge number of clients that are behind NAT and share an IP.
2. Add a configuration flag so that this behavior is configurable and would not dump client IPs into logs by default

ggarber:
@drmanu011 can you review the feedback received?

drmanu011:
Yes, it is easier to use a function rather than duplicating the code.

eakraly:
There is a new mechanism to rate limit invalid traffic enable through `--unauthorized-ratelimit`
Please give it a try
Logging invalid IPs and then blocking them through firewall might be counter productive as it will reduce performance of the server and increase impact of attacks

PR #1663, opened 2025-03-28
change for incorrect handling of TURN_NO_TLS environment variable

leonidgorbachev:
In version 4.5.1.3, there was a handling of TURN_NO_TLS environment variable and the TURN_xxx environment variables can be set in the following manner:
      **export TURN_NO_TLS=1**
      export TURN_NO_PQ=1
      export TURN_NO_MYSQL=1
      export TURN_NO_MONGO=1
      export TURN_NO_SQLITE=1
      ./configure

In versions 4.5.2 and later, handling of TURN_NO_TLS was removed, and with this TURN_NO_TLS is set as-is OSCFLAGS:
https://github.com/coturn/coturn/blob/831b7245aa03a230178a6b617b28264def4047d7/configure#L1021
And now, to set no-tls properly during compile time, we have to do the following:
      **export TURN_NO_TLS=-DTURN_NO_TLS**
      export TURN_NO_PQ=1
      export TURN_NO_GCM=1

Otherwise, if leave previous approach, it obviously causes an error during compile time:
```
    cc: error: 1: linker input file not found: No such file or directory
```

With this PR i would like to restore consistency of environment variables that used for ./configure


eakraly:
Thank you for the fix @leonidgorbachev !

Issue #1661, opened 2025-03-24
Client library support for Mbed-TLS


Luminoth:
I'm working on a project currently that is using [Mbed-TLS](https://github.com/Mbed-TLS/mbedtls) instead of OpenSSL in a client context and I'm wondering if it would be acceptable to add support to the Coturn client library for building against that. I've poked around the source a bit and it seems like it should be reasonably easy, aside from the OAuth support, which I'm wondering if that might make sense as its own separate library?

I'm happy to put up PRs for this work, assuming it's deemed reasonable to move forward on. Happy to work with and engage as much as needed for this.

eakraly:
Hi @Luminoth !
Sure why not.

Luminoth:
Awesome! The first PR is up to split out the client OAuth piece.

PR #1664, opened 2025-03-28
Split client OAuth into its own library

Luminoth:
This pulls the OAuth code out of the client library and into its own library (clientoauth). The main drive for this comes from https://github.com/coturn/coturn/issues/1661, mainly because for that project we don't need client OAuth (at the moment) and splitting it allows for the client lib to build against Mbed-TLS without having to convert over all of the OAuth pieces.

This is a breaking change obviously for any projects using the OAuth pieces of the client lib. I think it's pretty easy to update, just include the new header `ns_turn_oauth.h` and link against the `clientoauth` library. I also didn't see any tests or anything that need to be run, so if there's anything in particular I need to test out please let me know!


ggarber:
Thank you for this work @Luminoth .   Is it an option to keep oauth funcions inside the existing "client" module but put them behind a "ifdef OAUTH_SUPPORT" so that you can disable that support in your builds?   I´m a bit reluctant to add more projects/modules if we can avoid it.   What do you think?

ggarber:
@Luminoth can you take a look at my comment if you have some time? thank you

Luminoth:
Sooooo, I've actually be laid off from the role I was in for this. I think it's still useful tho to have that in here, I'll set aside some time Monday to revisit this. I think I still also have the changes in another branch after this to add the mbedtls support from the original thread.

Luminoth:
Sorry, I got delayed on looking back at this for a sec. Looking at how things are setup, the oauth, relay, and rfc5769 apps all just sort of use the oauth functions. Is there a way you're imaging that to be handled? The rfc5769 app seems like it might be easy to re-use the define but the relay and oauth apps seem pretty coupled to it. Would you just not build them at all maybe if the define was off?

Depending on how deep this particular rabbit hole goes I might reevaluate how worthwhile this work actually is right now.

ggarber:
Sorry to hear about the lay offs and thank you for keeping contributing @Luminoth.  I was thinking something like this :

bool convert_oauth_key_data(...) {
#ifndef OAUTH_DISABLED
....
#else
LOG("Oauth support not included")
return false;
#endif
}

Is that possible?  Thank you.

Luminoth:
Ok, yeah that makes sense. I might end up abandoning this PR then and just redoing it since that's gonna be way less overall changes to the code.

Luminoth:
Closing this in favor of https://github.com/coturn/coturn/pull/1713

PR #1671, opened 2025-05-01
[Quick Fix] Fix Repeated Line in Example turnserver.conf

stefankiesz:
The `#allocation-default-address-family="ipv4"` line is repeated twice in the example config, changed the second one to be `"ipv6"` which I assume it was intended to be.


eakraly:
Thanks for spotting this! Actually, this is just a duplicate line

stefankiesz:
No problem! I updated it to just remove the whole line.

PR #1692, opened 2025-05-29
Disable deprecated tls ssl

jonesmz:
Formally deprecated by IETF in March 2021: https://www.rfc-editor.org/rfc/rfc8996.html


eakraly:
Actually I agree with @ggarber 
Going to revert and edit @jonesmz PR to make TLS1 and 1.1 optional (default off)

PR #1693, opened 2025-05-30
Make older TLS versions optional

eakraly:
TLSv1 and TLSv1.1 can be enabled using `--tlsv1` and `--tlsv1_1` arguments accordingly
That assumes openssl version being used has these versions enabled (which as of openssl-3.5 is not by default)


ggarber (approved):
thank you!

PR #1695, opened 2025-05-31
feat: Add option to disable error reasons in STUN responses

ggarber:
This commit introduces a new configuration option, `send-error-reason`. When enabled, the TURN server will send STUN error responses with an reason phrase.   By default it wont send the error reason so it is a change in the default behaviour.

Changes include:
- Added `send_error_reason` to the `turn_params_t` struct.
- Modified `mainrelay.c` to parse the `--no-error-reason` command-line option and corresponding configuration file setting.
- Updated `ns_turn_server.c` to check this flag and conditionally omit the reason phrase when constructing error responses.
- Added the new option to the example `turnserver.conf` file.
- Conceptually outlined test cases for this new functionality.


ggarber:
Still some work pending, i´ll try to remove the reason completely instead of leaving it empty.   I saw that in Cloudflare implementation so it looks safe.

eakraly:
implemented in #1829

PR #1702, opened 2025-06-10
guarantee no oob write in strncat

redraincatching:
based on [this codescanning alert](https://github.com/redraincatching/coturn/security/code-scanning/166)

guarantees that the `turn_params.cipher_list` will be null-terminated, and that the call to strncpy cannot attempt to access out-of-bounds memory


ggarber:
bugbot run

PR #1706, opened 2025-06-20
removed unnecessary null check

redraincatching:
removing an unnecessary null check as raised in [this code scan](https://github.com/redraincatching/coturn/security/code-scanning/179)

the variable is confirmed to not be null in an outer loop guard

also changed the name of the variable in `rtcp_map_put`'s function declaration to match that used in its function definition


eakraly (approved):
Thank you!

PR #1709, opened 2025-06-26
refactored tri-state to bools

redraincatching:
refactored random tri-state to use two random booleans for clarity


ggarber:
thank you very much!

PR #1710, opened 2025-06-26
reduce calls to random

redraincatching:
implemented change suggested in TODO to speed up aes key generation without, hopefully, negatively impacting the overall randomness of the function

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -1773,7 +1773,7 @@ static void generate_aes_128_key(char *filePath, unsigned char *returnedKey) {
     key[i] = (i < 8) ? (random_value_0 >> (i * 8)) & 0xff : (random_value_1 >> ((i - 8) * 8)) & 0xff;
   }
 
-  FILE *fptr = fopen(filePath, "w");
+  FILE *fptr = fopen(filePath, "wb");
   if (!fptr) {
     return;
   }


src/apps/relay/mainrelay.c
```diff
@@ -1801,7 +1798,7 @@ unsigned char *base64decode(const void *b64_decode_this, int decode_this_many_by
   int decoded_byte_index = 0;                     // Index where the next base64_decoded byte should be written.
   while (0 < BIO_read(b64_bio, base64_decoded + decoded_byte_index, 1)) { // Read byte-by-byte.
     decoded_byte_index++; // Increment the index until read of BIO decoded data is complete.
-  }                       // Once we're done reading decoded data, BIO_read returns -1 even though there's no error.
+  } // Once we're done reading decoded data, BIO_read returns -1 even though there's no error.
```
redraincatching, line 1801: clang-format change

ggarber:
Thank you very much @redraincatching 🙇 

Why is the change showing in GH as if the whole file was changed?  Can you fix that?

Also review copilot suggestion, it looks like a bug.

redraincatching:
@ggarber the whole file being changed was windows line endings, my bad, have that resolved now.

the copilot comment is irrelevant, as i am not using the masked values as numeric values, i'm just using them to generate a boolean, so no right-shift is required.

ggarber:
> @ggarber the whole file being changed was windows line endings, my bad, have that resolved now.
> 
> the copilot comment is irrelevant, as i am not using the masked values as numeric values, i'm just using them to generate a boolean, so no right-shift is required.

but wont "uint8_t random_val_1 = r & 0x0000ff00;" be always 0 because of the cast ot uint8_t ?

redraincatching:
@ggarber i think i fully forgot which one of my pull requests this was, you're right

that should be fixed now

ggarber:
Thank you very much @redraincatching .

I'm reviewing the code now and I'm not sure the new code is not degrading the entropy/randomness of the key generation.

With "random_val_2 % 26" there are some numbers less probable than others (I think 25 is less probable than 1).   Is that ok?

With turn_random*() % 26 that problem is much lower because turn_random() is a very big number instead of just a 0-255 value like after the change.

What do you think?   I'm not an expert on this but I think we shouldn't introduce any security degradation just to save some perfomrance in a function call that happens only once during startup.

cc @eakraly

jonesmz:
> With turn_random*() % 26 that problem is much lower because turn_random() is a very big number instead of just a 0-255 value like after the change.

The expression ` val % 26` always returns a value between 0 and 25, so no matter what the original value of `val` is, from a bit-wise randomness point of view, only the bits that might show up in a value of 25 or less would matter, and anything in excess of that would be discarded.

Of course, this assumes that the random values returned are quality randomness, and that the function generating the random number is properly filling the entire size_t.

ggarber:
> > With turn_random*() % 26 that problem is much lower because turn_random() is a very big number instead of just a 0-255 value like after the change.
> 
> The expression ` val % 26` always returns a value between 0 and 25, so no matter what the original value of `val` is, from a bit-wise randomness point of view, only the bits that might show up in a value of 25 or less would matter, and anything in excess of that would be discarded.
> 
> Of course, this assumes that the random values returned are quality randomness, and that the function generating the random number is properly filling the entire size_t.

But the chance of getting 25 (versus getting 1 for example) is higher with the previous implementation than with the current one.   Does that affect the quality of the key?

jonesmz:
> > > With turn_random*() % 26 that problem is much lower because turn_random() is a very big number instead of just a 0-255 value like after the change.
> > 
> > 
> > The expression ` val % 26` always returns a value between 0 and 25, so no matter what the original value of `val` is, from a bit-wise randomness point of view, only the bits that might show up in a value of 25 or less would matter, and anything in excess of that would be discarded.
> > Of course, this assumes that the random values returned are quality randomness, and that the function generating the random number is properly filling the entire size_t.
> 
> But the chance of getting 25 (versus getting 1 for example) is higher with the previous implementation than with the current one. Does that affect the quality of the key?

I am not understanding how you are arriving at the probability of a 25 being higher or lower. Can you walk me through your scenario where this would be the case?

ggarber:
we are using an uint8_t so the value is 0-255, right?

in the range 0-255 there are more numbers that % 26 give 1 than numbers that % by 26 give 25.

Screenshot 2025-07-02 at 11 39 52

Generated with this code:
```
from collections import Counter

remainders = [i % 26 for i in range(0, 256)]
count = Counter(remainders)

for i in range(26):
    print(f"{i}: {count[i]}")
```

jonesmz:
Ahhhhh, I understand your concern now.

So the real answer is that this function is literally just a hexadecimal string generator.

Since the transform from an arbitrary byte sequence to hexadecimal is just 8bit -> 16bit, we can generate 8bytes of random bytes and run that through a lookup table to arrive at 16bytes of output.

The use of one call to rand per loop, as in this PR, or 2 calls per loop in the original implementation can be replaced with a total of 2 calls for 32 bit platforms, and 1 call on 64bit platforms.

redraincatching:
@ggarber if we're just generating a hex string, this newest implementation should be faster than the original, and also avoids the issue of uneven distribution of values, increasing the overall randomness

ggarber:
I'm sorry but I don't think this code is generating a 128bit key.    It is 16 bytes of only 4 bits of randomness each.   You can not generate a random key of 128bit with a single random number of 64 bits.

The implementation should be something like this:  
    for (int i = 0; i < 4; ++i) {
      uint32_t rnd = turn_random();
      memcpy(&key_buffer[i * 4], &rnd, 4);
    }
    for (int i = 0; i < 16; ++i) {
        fprintf(fptr, "%02x", key_buffer[i]);
    }      
    
(code contributed by chatgpt)

Thank you for so many iterations.

jonesmz:
The original implementation was not generating 128 bits of randomness, its generating a 16byte hex string.

Is the output of the function intended to be true random values, or is it supposed to be hex?


Edit: you can see at the end of the function that only 16 bytes are written to the file, in both this implementation and the original one.

If we want 128 bits of true randomness written into the file, then formatting it as hex means it needs to write 32 bytes (256 bits) into the file.

ggarber:
> The original implementation was not generating 128 bits of randomness, its generating a 16byte hex string.

you are right, the original implementation looks also wrong :(

> Is the output of the function intended to be true random values, or is it supposed to be hex?

Based on the description of the flag:
"	-x, --key-path			Generates a 128 bit key into the given path.\n"
And how it is used later on in the code (generate_aes_128_key, encrypt_aes_128, decrypt_aes_128) I think we should just store the key as 16 random bytes (not hex) in the file.

> Edit: you can see at the end of the function that only 16 bytes are written to the file, in both this implementation and the original one.
> 
> If we want 128 bits of true randomness written into the file, then formatting it as hex means it needs to write 32 bytes (256 bits) into the file.

redraincatching:
@ggarber I've now refactored it to generate 128 bits of purely random noise, instead of the hex values that were being generated.  That does seem more in line with how the flag is described, and how the function is used. 

The only downside I can think of is that the key no longer uses only printable characters, but if that's an issue, the `base64encode` function immediately above this in the same file is probably a better alternative, since the hex string used is not 128 bits of randomness, as @jonesmz pointed out.

src/apps/relay/mainrelay.c
```diff
@@ -1753,28 +1754,25 @@ void encrypt_aes_128(unsigned char *in, const unsigned char *mykey) {
...
-      continue;
-    case 1:
-      key[i] = (turn_random() % 26) + 65;
-      continue;
-    default:
-      key[i] = (turn_random() % 26) + 97;
-      continue;
-    }
+    // store the 128 random bits in the key array
+    key[i] = (i < 8) ? (random_value_0 >> (i * 8)) & 0xff : (random_value_1 >> ((i - 8) * 8)) & 0xff;
```
ggarber, line 1773: I think & 0xff is not needed but LGTM.  Thank you!

ggarber (approved):

src/apps/relay/mainrelay.c
eakraly: would splitting into 2 loops (i<8 and i>=8) perform even better because no ifs?
Or even unrolling the loops
Looking at it even farther - this is inversing order of bytes (like ntohl) - maybe we can use something like it? I'm sure there are faster implementation (like intrinsics)

ggarber:
I think i'm going to merge it (I just fixed the latest copilot suggestion) unless somebody disagrees. @eakraly @redraincatching @jonesmz   Thank you everybody for helping!

PR #1737, opened 2025-08-16
mainreplay: fix leak in decrypt_aes_128()

brandb97:
In decrypt_aes_128() at src/apps/replay/mainreplay.c, it calls base64decode() to allocates memory in encryptedText, but forgets to free encryptedText in the end of this function. Add free() after finished using encryptedText.


eakraly:
Nice find @brandb97

tuyuribr (approved):

PR #1744, opened 2025-08-28
address possible null pointer dereferences

redraincatching:
# addressing all remaining code scanning instances of warning C6011, null pointer dereference

this pr aims to address more static code analyser warnings, specifically null pointer dereferences. the majority of changes are solely to quieten the analyser, as `malloc` and `calloc` are unlikely to fail, but this should at least lead to the code analysis being more readable and usable.

where functions addressed had existing failure strategies, they were used, however some functions will now silently fail rather than attempting to dereference a null pointer. if there is a preferred solution in these cases, i will be happy to implement it.

---

this is an extension of [this pull request](https://github.com/coturn/coturn/pull/1729)

### DIFF
--- a/src/apps/relay/http_server.c
+++ b/src/apps/relay/http_server.c
@@ -220,6 +220,11 @@ struct http_request *parse_http_request(char *request) {
 
     ret = (struct http_request *)calloc(sizeof(struct http_request), 1);
 
+    if (ret == NULL) {
+      TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "%s: failure in call to calloc \n", __FUNCTION__);
+      return NULL;
+    }
+
     if (strstr(request, "GET ") == request) {
       ret->rtype = HRT_GET;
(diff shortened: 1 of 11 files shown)


src/apps/relay/http_server.c
```diff
@@ -220,6 +220,11 @@ struct http_request *parse_http_request(char *request) {
 
     ret = (struct http_request *)calloc(sizeof(struct http_request), 1);
 
+    if (ret == NULL) {
+      TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "!!! %s: failure in call to calloc !!!\n", __FUNCTION__);
```
eakraly, line 224: Let's not add `!!!` to error messages - they do not mean anything
ERROR log is enough to signal an error and importance of the log message

eakraly (changes requested):

src/apps/relay/http_server.c
redraincatching: very fair point, i've removed them now

PR #1751, opened 2025-09-14
Fix memory leak using libevent

### LINKED ISSUE #1750: Memory leak in turn_event_base_new() due to missing event_config_free()
In file apputils.c the code should be modified as shown below. This problem was flagged by valgrind in my tests.

struct event_base *turn_event_base_new(void)
{
	struct event_config *cfg = event_config_new();

	event_config_set_flag(cfg,EVENT_BASE_FLAG_EPOLL_USE_CHANGELIST);

#ifdef IMPROVED_CODE
	struct event_base *base = event_base_new_with_config(cfg);
	event_config_free(cfg);  // Free the config after use to make valgrind happy
	return base;
#else //existing code
	return event_base_new_with_config(cfg);
#endif
}

eakraly:
Fixes #1750

### DIFF
--- a/src/apps/common/apputils.c
+++ b/src/apps/common/apputils.c
@@ -927,8 +927,8 @@ static char *_WTA(__in wchar_t *pszInBuf, __in int nInSize, __out char **pszOutB
     return NULL;
   } else {
     if (pszOutBuf != NULL) {
-    (*pszOutBuf)[*pnOutSize - 1] = '\0';
-    return *pszOutBuf;
+      (*pszOutBuf)[*pnOutSize - 1] = '\0';
+      return *pszOutBuf;
     } else {
       return NULL;
     }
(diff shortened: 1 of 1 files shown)


jonesmz:
This change seems to be removing the const keyword in several places without changing any program logic related to the formerly const variable.

src/apps/common/apputils.c
```diff
@@ -919,19 +919,12 @@ static char *_WTA(__in wchar_t *pszInBuf, __in int nInSize, __out char **pszOutB
   // and we have to add space to the allocation ourselves.
   (*pnOutSize)++;
   *pszOutBuf = malloc(*pnOutSize * sizeof(char));
-  if (!pszOutBuf) {
```
redraincatching, line 922: is there a reason for removing these checks? it'll cause unnecessary noise in the code scanner

eakraly:
ehh bad rebase?

PR #1752, opened 2025-09-17
change port identifiers to use uint16_t

redraincatching:
based on the ideas originally developed in [this pr](https://github.com/coturn/coturn/pull/1535) by @WHYHD

### DIFF
--- a/src/apps/natdiscovery/natdiscovery.c
+++ b/src/apps/natdiscovery/natdiscovery.c
@@ -330,8 +330,7 @@ static int run_stunclient_lifetime(int timer, ioa_addr *local_addr, ioa_addr *re
   init_socket(&udp_fd2, local_addr, 0, remote_addr);
   response_port = addr_get_port(reflexive_addr);
 
-  ret = stunclient_send(udp_fd2, local_addr, local_port, remote_addr, change_ip, change_port, padding,
-                        response_port);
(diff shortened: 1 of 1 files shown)


src/apps/natdiscovery/natdiscovery.c
```diff
@@ -508,7 +515,8 @@ static int run_stunclient(ioa_addr *local_addr, ioa_addr *remote_addr, ioa_addr
 
   init_socket(&udp_fd, local_addr, *local_port, remote_addr);
 
-  ret = stunclient_send(&buf, udp_fd, local_addr, local_port, remote_addr, change_ip, change_port, padding, -1);
+  ret = stunclient_send(&buf, udp_fd, local_addr, (uint16_t *)local_port, remote_addr, change_ip, change_port, padding,
```
eakraly, line 518: Why not change the argument to the function instead?

src/apps/natdiscovery/natdiscovery.c
```diff
@@ -525,15 +533,17 @@ static int run_stunclient_hairpinning(ioa_addr *local_addr, ioa_addr *remote_add
...
```
eakraly, line 536: Same... instead of casting should change the function argument

eakraly (changes requested):
almost there - a few more places to change the type

PR #1768, opened 2025-10-20
Allow faster packet validation on listener threads

eakraly:
Add CLI flag to enable early packet validation
Drop packets that do not pass basic STUN command/channel or DTLS parsing

Before this change, no validation on packets were done and they were passed (through libevent queue) to relay thread pool. Relay thread would, for a new source, allocate a new SS (18KB) which will only be released after 1s of no traffic, and then do the validation.

So with old code, invalid packet would have extra:
- Queuing
- Processing on a different thread
- Memory allocation of 18KB

Assuming DDoS attack is spoofing IPs it reduces processing capacity dramatically.
Testing possible by:
```
hping3 -2 -p 3478 -d 2 -rand-source --flood turn_ip
```

which floods `turn_ip:3478` with packets of size 2 from random sources. Size 2 is especially bad case - the packet is obviously invalid (too short) but still goes through a long process of queuing, thread switching, memory allocation and only then validation (and then memory cleanup etc). In worst cases, memory is never cleaned up because sources repeat.

### DIFF
--- a/src/apps/relay/dtls_listener.c
+++ b/src/apps/relay/dtls_listener.c
@@ -64,7 +64,11 @@ typedef uint16_t in_port_t;
 
 #define MAX_SINGLE_UDP_BATCH (16)
 
+#if !defined(WINDOWS)
 _Thread_local uint32_t packetcounter = 0;
+#else
+static uint32_t packetcounter = 0;
+#endif
 
 struct dtls_listener_relay_server_info {
   char ifname[1025];


jonesmz:
Can you elaborate on why you're making this a commandline option, instead of always enabling the basic validation?

eakraly:
> Can you elaborate on why you're making this a commandline option, instead of always enabling the basic validation?

It should go through some soak testing - new features should be flagged IMHO. In this way people do not get nasty surprises that something new is on without any flag change.

But I see your point - I can suggest to invert the flag: in this way you get it on by default but can turn it off for whatever reason. In the release after that I can remove the flag completely.

WDYT @jonesmz ?

jonesmz:
From my perspective, coturn is performant enough that my usage of it doesn't really need to run additional instances until we see many many many thousands of concurrent connections. So a small additional per-packet cost to validate correctness isn't much of a concern.

On the other hand, running a public facing service on the internet that my customers rely on for connectivity, means that I'd rather see mitigations to protect against denial of service attacks.

So for that, I agree with your suggestion, better to have the protection by default.

src/apps/relay/dtls_listener.c
```diff
@@ -130,6 +133,33 @@ int get_dtls_version(const unsigned char *buf, int len) {
   return 0;
 }
 
+static size_t print_packet_txt2pcap(uint64_t now, uint8_t *payload, size_t payload_length, uint8_t *txt2pcap,
+                                    size_t txt2pcap_length) {
+
+  size_t index = 0;
+  int64_t remaining = now % (24 * 60 * 60 * 1000);
```
jonesmz: unsigned to signed conversion. 

Is there any benefit to converting to signed, instead of staying unsigned?

src/apps/relay/dtls_listener.c
```diff
@@ -130,6 +133,33 @@ int get_dtls_version(const unsigned char *buf, int len) {
...
+  int hours = remaining / (60 * 60 * 1000);
+  remaining = remaining % (60 * 60 * 1000);
+  int minutes = remaining / (60 * 1000);
+  remaining = remaining % (60 * 1000);
+  int seconds = remaining / 1000;
+  int ms = remaining % 1000;
+
+  snprintf((char *)(txt2pcap + index), txt2pcap_length - index, "%02d:%02d:%02d.%03d", hours, minutes, seconds, ms);
```
jonesmz: At this point in the function `index` is always zero.

it doesn't get changed to non-zero until line 150.

Maybe move the declaration of index to line 150?

src/apps/relay/dtls_listener.c
```diff
@@ -130,6 +133,33 @@ int get_dtls_version(const unsigned char *buf, int len) {
...
+  index += sizeof("00:00:00.000") - 1;
+  snprintf((char *)(txt2pcap + index), txt2pcap_length - index, " 0000");
+  index += sizeof(" 0000") - 1;
+
+  for (size_t i = 0; i < payload_length; i++) {
+    snprintf((char *)(txt2pcap + index), txt2pcap_length - index, " %02x", payload[i]);
+    index += 3;
+  }
+  snprintf((char *)(txt2pcap + index), txt2pcap_length - index, " # STUN_PACKET ");
+  index += sizeof(" # STUN_PACKET ");
```
jonesmz: Instead of using sizeof, wouldn't it be better to use the return of snprintf?

https://en.cppreference.com/w/c/io/fprintf.html

> 4) number of characters (not including the terminating null character) which would have been written to buffer if bufsz was ignored, or a negative value if an encoding error (for string and character conversion specifiers) occurred.

That way future changes to the function can't screw up by only changing the string in one place.

Issue #1765, opened 2025-10-19
Potential NULL Pointer Dereference found in `set_option`


MicroMilo:
Similar to other issues, I have found a potential NULL Pointer Dereference. The file location is:

> https://github.com/coturn/coturn/blob/242467faba87b434fa901841ec72c4fe3da79d5d/src/apps/relay/mainrelay.c#L1879

The `set_option` function is very large, but in some branches, the `value` variable is not checked (at least 30+ similar problems).
Take an example of the following code, I have made some comments to help understand. 

```c
static void set_option(int c, char *value) {
  if (value && value[0] == '=') {  // ============ the value variable could be NULL.
    TURN_LOG_FUNC(TURN_LOG_LEVEL_WARNING,
                  "WARNING: option -%c is possibly used incorrectly. The short form of the option must be used as "
                  "this: -%c <value>, no \'equals\' sign may be used, that sign is used only with long form options "
                  "(like --user=<username>).\n",
                  (char)c, (char)c);
  }

  switch (c) {
  case 'K':
    if (get_bool_value(value)) {
      turn_params.allocation_default_address_family = ALLOCATION_DEFAULT_ADDRESS_FAMILY_KEEP;
    }
    break;
  case 'A':
    if (value && strlen(value) > 0) { // ============ the value cound be NULL also.
      if (*value == '=') {
        ++value;
      }
      if (!strcmp(value, "ipv6")) {
        turn_params.allocation_default_address_family = ALLOCATION_DEFAULT_ADDRESS_FAMILY_IPV6;
      } else if (!strcmp(value, "keep")) {
        turn_params.allocation_default_address_family = ALLOCATION_DEFAULT_ADDRESS_FAMILY_KEEP;
      } else if (!strcmp(value, "ipv4")) {
        turn_params.allocation_default_address_family = ALLOCATION_DEFAULT_ADDRESS_FAMILY_IPV4;
      } else {
        TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "ERROR: invalid allocation_default_address_family parameter\n");
      }
    } else {
      TURN_LOG_FUNC(TURN_LOG_LEVEL_ERROR, "ERROR: invalid allocation_default_address_family parameter\n");
    }
    break;
  case SERVER_NAME_OPT: // =========== when the value is NULL, however, it is not checked！
    STRCPY(turn_params.oauth_server_name, value);
    break;
// ...
  case 'i':
[...]

Issue #1763, opened 2025-10-14
Potential Null Pointer Dereference found due to Unreasonable Check


MicroMilo:
In the function `turnports_allocate`, due to the Unreasonable Check, I think there is a possibility of causing NPD.
In language C, the priority of operations of `->` is greater than `&`.
The code in the function checks if the variable `tp` NULL; However, if it is NULL, the `NPD` has happened due to the `->` operation.

#### case 1

> https://github.com/coturn/coturn/blob/242467faba87b434fa901841ec72c4fe3da79d5d/src/apps/relay/turn_ports.c#L165

```c
int turnports_allocate(turnports *tp) {

  int port = -1;

  TURN_MUTEX_LOCK(&tp->mutex);  // ============ NPD here

  if (tp) {  // ======== post check here.
    //.. 
  }

  TURN_MUTEX_UNLOCK(&tp->mutex);

  return port;
}
```

#### case 2

The same thing happens in the function `turnports_release`

> https://github.com/coturn/coturn/blob/242467faba87b434fa901841ec72c4fe3da79d5d/src/apps/relay/turn_ports.c#L206

```c
void turnports_release(turnports *tp, uint16_t port) {
  TURN_MUTEX_LOCK(&tp->mutex);  // ========== NPD here
  if (tp && port >= tp->range_start && port <= tp->range_stop) {  // ======== post check here.
    const uint16_t position = (uint16_t)(tp->high & 0x0000FFFF);
    if (is_taken(tp->status[port])) {
      tp->status[port] = tp->high;
      tp->ports[position] = port;
      ++(tp->high);
    }
  }
  TURN_MUTEX_UNLOCK(&tp->mutex);
}
```

#### case 3

Something more terrible happens in the function `turnports_allocate_even`, it checks if `tp` is NULL first; however, when `tp` is NULL, it will execute the code of `TURN_MUTEX_UNLOCK(&tp->mutex);`, which will cause NPD directly.

> https://github.com/coturn/coturn/blob/242467faba87b434fa901841ec72c4fe3da79d5d/src/apps/relay/turn_ports.c#L219

```c
int turnports_allocate_even(turnports *tp, int allocate_rtcp, uint64_t *reservation_token) {
  if (tp) {  // =============== pre check here
    TURN_MUTEX_LOCK(&tp->mutex);
    //..
  }

  TURN_MUTEX_UNLOCK(&tp->mutex); // ============ NPD here

  return -1;
}
```

[...]

Issue #1762, opened 2025-10-14
Potential Null Pointer Dereference found in `post_parse`


MicroMilo:
In the function `post_parse`, `post_data` is allocated, followed by a check to see if the memory is allocated successfully.
However, why does the variable `list` miss the check?
So I think it is more possible to case NPD here.

> https://github.com/coturn/coturn/blob/242467faba87b434fa901841ec72c4fe3da79d5d/src/apps/relay/http_server.c#L112

```c
static struct headers_list* post_parse(char* data, size_t data_len) {
    //..
    char* post_data = (char*)calloc(data_len + 1, sizeof(char));
    if (post_data != NULL) {
        //..
        struct headers_list* list = (struct headers_list*)calloc(sizeof(struct headers_list), 1);  // =========== allocated here.
        //..
        list->keys = (char**)realloc(list->keys, sizeof(char*) * (list->n + 1)); // =========== potential NPD here
        list->keys[list->n] = strdup(key);
        list->values = (char**)realloc(list->values, sizeof(char*) * (list->n + 1));
        list->values[list->n] = value;
        ++(list->n);
    }
    //..
}
```

ggarber:
If we run out of memory and calloc starts failing coturn will probably fail in many other places.     coturn is not handling those cases of "system failure" gracefully.

I think it is not a critical issue but we are definitely open to accept a PR to fix cases like this one, specially if/when they are easy to fix and not have side effects.    Thank you very much @MicroMilo for this report and the others similar to this one, very appreciated 🙇

MicroMilo:
Ok, I will try to do that.

PR #1769, opened 2025-10-21
Fix bug caused by missing return value check

### LINKED ISSUE #1766: Potential NULL Pointer Dereference or Divede Zero found in `client_read`
By scanning code, I found the `clmessage_length` is configurable in`main.c`, which is in the `main` function.

> https://github.com/coturn/coturn/blob/242467faba87b434fa901841ec72c4fe3da79d5d/src/apps/uclient/mainuclient.c#L300

However, a potential bug could be triggered in the function `client_read`

> https://github.com/coturn/coturn/blob/242467faba87b434fa901841ec72c4fe3da79d5d/src/apps/uclient/uclient.c#L647

### case 1: NULL Pointer Dereference

> https://github.com/coturn/coturn/blob/242467faba87b434fa901841ec72c4fe3da79d5d/src/apps/uclient/uclient.c#L735

As for the stun_attr_get_value, it could return NULL, which may happen when the `rlen` is zero (less than 1) and `clmessage_length` is zero.

- `rlen` is the length of the `DATA` which could be sent by the user.

> https://github.com/coturn/coturn/blob/242467faba87b434fa901841ec72c4fe3da79d5d/src/client/ns_turn_msg.c#L1286

```c
        int rlen = stun_attr_get_len(sar);
        applen = rlen;
        if (rlen != clmessage_length) {
          TURN_LOG_FUNC(TURN_LOG_LEVEL_INFO, "ERROR: received DATA message has wrong len: %d, must be %d\n", rlen,
                        clmessage_length);
          tot_recv_bytes += applen;
          return rc;
        }
        const uint8_t *data = stun_attr_get_value(sar); // ========= stun_attr_get_value could return NULL
        memcpy(&mi, data, sizeof(message_info));  
```

### case 2: Divide by Zero

Easy to see, when `clmessage_length` is zero, Divide by Zero may happen.

> https://github.com/coturn/coturn/blob/242467faba87b434fa901841ec72c4fe3da79d5d/src/apps/uclient/uclient.c#L691

```c
    if (is_tcp_data) {
      if ((int)elem->in_buffer.len == clmessage_length) {
        memcpy(&mi, (elem->in_buffer.buf), sizeof(message_info));
        miset = true;
      } else {
        /* TODO: make a more clean fix */
        buffers = (int)elem->in_buffer.len / clmessage_length;  // =========== the zero value never be checked!
      }
    }
```

MicroMilo:
Fix https://github.com/coturn/coturn/issues/1766

### DIFF
--- a/src/apps/common/apputils.c
+++ b/src/apps/common/apputils.c
@@ -1066,6 +1066,9 @@ void set_execdir(void) {
 #endif
   if (_var && *_var) {
     _var = strdup(_var);
+    if (!_var) {
+      return;
+    }
     char *edir = _var;
     if (edir[0] != '.') {
       edir = strstr(edir, "/");
@@ -1135,6 +1138,9 @@ char *find_config_file(const char *config_file) {
         const size_t dirlen = strlen(config_file_search_dirs[i]);
         size_t fnsz = sizeof(char) * (dirlen + cflen + 10);
         char *fn = (char *)malloc(fnsz + 1);
+        if (!fn) {
+          break;
+        }
         strncpy(fn, config_file_search_dirs[i], fnsz);
         strncpy(fn + dirlen, config_file, fnsz - dirlen);
         fn[fnsz] = 0;
@@ -1149,6 +1155,9 @@ char *find_config_file(const char *config_file) {
           const size_t celen = strlen(c_execdir);
           fnsz = sizeof(char) * (dirlen + cflen + celen + 10);
           fn = (char *)malloc(fnsz + 1);
+          if (!fn) {
+            break;
+          }
           strncpy(fn, c_execdir, fnsz);
           size_t fnlen = strlen(fn);
           if (fnlen < fnsz) {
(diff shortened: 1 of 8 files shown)


MicroMilo:
There are also some similar bugs need to fix.
I'll fix it as soon as I can.

eakraly:
Hi @MicroMilo 
Thank you for the PR

While some of the changes make sense (clmessage length and division by 0) - the rest of return value checks are actually bad IMHO
There is no backup plan or recovery from failure to allocate memory and it is good that the app crashes
If malloc/calloc/realloc fail - there is absolutely no point to try and continue running. Moreover, it is harmful as it hides the real issue (if the app actually runs for some time after failure to allocate memory)
In some place, the fix includes logging the issue - which again needs memory allocated. In this case, best case scenario, the crash will show that memory allocation failed in logging - which is not true issue.

This is a generic issue with automatic scanners that warn about return value validation. In case of memory allocation - those do not have value.

To sum it up: I recommend against memory allocation validation changes - I'll review the rest after that.

Thanks again!

MicroMilo:
Thanks, @eakraly.
Thank you for your response. I think I know what you think about memory management.
I put forward this PR out of the discussion in https://github.com/coturn/coturn/issues/1762.

Leaving memory management aside, I still have some issues regarding null Pointers. Do you want me to request a PR?
for example: https://github.com/coturn/coturn/issues/1763, https://github.com/coturn/coturn/issues/1765

eakraly:
closing PR as discussed

PR #1779, opened 2025-12-05
Use bool to enable prometheus

eakraly:
`turn_params.prometheus` is bool but used as int throughout the code
`turn_params,prometheus_username_labels` is bool but in one place is set with 1

This PR changes 0 and 1 to false and true accordingly

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -2226,7 +2226,7 @@ static void set_option(int c, char *value) {
     STRCPY(turn_params.prometheus_path, value);
     break;
   case PROMETHEUS_ENABLE_USERNAMES_OPT:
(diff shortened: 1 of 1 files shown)


src/apps/relay/mainrelay.c
```diff
@@ -2226,7 +2226,7 @@ static void set_option(int c, char *value) {
     STRCPY(turn_params.prometheus_path, value);
     break;
   case PROMETHEUS_ENABLE_USERNAMES_OPT:
-    turn_params.prometheus_username_labels = 1;
+    turn_params.prometheus_username_labels = value ? true : false;
```
eakraly: > 
```suggestion
    turn_params.prometheus_username_labels = true;
```

PR #1789, opened 2025-12-27
Fix crashes while setting socket buffer size

eakraly:
Some places in code do not have access to the buffer size which result in crash which can be seen in tests
This PR removes the call to `set_ioa_socket_buf_size` from those places (which is redundant anyway)


jonesmz:
Why wouldn't the buffer size be known? Some more context would be helpful in evaluating this change.

PR #1794, opened 2026-01-18
Add session usage reporting callback to TURN database driver

mesibo:
We use Coturn in production at [mesibo](https://mesibo.com) and identified a gap in dbdriver: there's no callback to capture post-session metrics like duration, bandwidth consumed, timestamps, etc. 

We've made the following changes, which could be useful for Coturn. Would appreciate your consideration. These changes are tested in production environments at mesibo.

## Changes for reporting usage

Added a `report_usage` callback to the database driver interface, invoked at session termination. This allows drivers to capture and persist detailed usage statistics in their preferred storage (MySQL, PostgreSQL, etc.).

### 1. Database Driver Interface
**File:** `src/apps/relay/dbdrivers/dbdriver.h`

```c
typedef struct _turn_dbdriver_t {
    // ... existing callbacks
    void (*report_usage)(void *);
} turn_dbdriver_t;
```

### 2. Session Termination Hook
**File:** `src/apps/relay/ns_ioalib_engine_impl.c`

```c
void turn_report_session_usage(void *session, int force_invalid) {
    // ... existing code
    
    if(force_invalid) {
        const turn_dbdriver_t * dbd = get_dbdriver();
        if(dbd->report_usage)
            dbd->report_usage(session);
    }

   // ... existing code
   ss->received_packets = 0;
   ...
}
```

### DIFF
--- a/src/apps/relay/dbdrivers/dbd_mongo.c
+++ b/src/apps/relay/dbdrivers/dbd_mongo.c
@@ -1378,14 +1378,15 @@ static void mongo_disconnect(void) {
 
 //////////////////////////////////////////////////////////
 
-static const turn_dbdriver_t driver = {
-    &mongo_get_auth_secrets,   &mongo_get_user_key,   &mongo_set_user_key,   &mongo_del_user,
-    &mongo_list_users,         &mongo_list_secrets,   &mongo_del_secret,     &mongo_set_secret,
-    &mongo_add_origin,         &mongo_del_origin,     &mongo_list_origins,   &mongo_set_realm_option_one,
-    &mongo_list_realm_options, &mongo_auth_ping,      &mongo_get_ip_list,    &mongo_set_permission_ip,
-    &mongo_reread_realms,      &mongo_set_oauth_key,  &mongo_get_oauth_key,  &mongo_del_oauth_key,
-    &mongo_list_oauth_keys,    &mongo_get_admin_user, &mongo_set_admin_user, &mongo_del_admin_user,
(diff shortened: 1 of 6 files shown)


eakraly:
@mesibo thanks for the PR
I have a few follow up questions:
1. There is no implementation in any of the DB drivers. Do you mind contributing a single implementation?
2. This should be controlled by a flag so that it is conditionally enabled and not forced upon
Thanks!

mesibo:
We initially had a MySQL driver implementation of report_usage, but later refactored our approach, removed all db drivers, and built a REST API-based driver instead. We can contribute if it works for you.

On the conditional flag — not sure that's necessary. Like any other function in the dbdriver interface, the implementation can simply choose to ignore it if not needed.

On a broader note, bundling DB drivers into Coturn pulls in external dependencies (libmysqlclient, libpq, etc.) that add unnecessary build and maintenance overhead for everyone. A better approach would be loadable modules, so operators can plug in their own reporting backend without touching the core Coturn build.

Moreover, we felt that DB drivers don't work well with dynamic username generation — one of the reasons we moved away from them in production. We generate usernames not just based on time, but also on call ID to control and gate traffic. 
Happy to raise a separate issue to describe this in more detail.

eakraly:
@mesibo Thanks for the explanation - those make sense to me.
Feel free to make additional contributions and I will try to review them asap!

mesibo:
> @mesibo Thanks for the explanation - those make sense to me. Feel free to make additional contributions and I will try to review them asap!

New issue added, we can discuss  https://github.com/coturn/coturn/issues/1815

PR #1809, opened 2026-02-20
OpenSSL: migrate to modern API for DH param

eakraly:
Use openssl-3.0 apis


eakraly:
Blocked by https://github.com/google/oss-fuzz/pull/14982

PR #1817, opened 2026-02-24
Restore support for OpenSSL >= 1.1.1 (for AlmaLinux/RHEL/Rocky Linux 8)

robert-scheck:
Conditionalize changes from commit 4c674289a8a288dd310e87490334684b05d61381 to support building using OpenSSL 1.1.1 in AlmaLinux/RHEL/Rocky Linux 8. While OpenSSL 3.5 is in EPEL 8 it unfortunately can not be used, because other (to be linked) libraries (libevent_openssl, libmariadb and libpq) in RHEL 8 are built and linked against OpenSSL 1.1.1 from RHEL 8.

### DIFF
--- a/src/apps/relay/mainrelay.c
+++ b/src/apps/relay/mainrelay.c
@@ -3503,8 +3503,12 @@ static void adjust_key_file_names(void) {
     adjust_key_file_name(turn_params.dh_file, "DH key", 0);
   }
 }
-static EVP_PKEY *get_dh566(void) {
 
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+static EVP_PKEY *get_dh566(void) {
+#else
+static DH *get_dh566(void) {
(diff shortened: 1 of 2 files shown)


robert-scheck:
@eakraly, any chance to get this merged – even you may have mixed feelings about it?

eakraly:
I understand the desire to have it in  but I've intentionally removed it - openssl 1.1.1 is fully deprecated for almost 3 years. But mostly, it creates maintenance overhead.
I can have as a standalone patch that can be applied to the code to enable support - if that helps

robert-scheck:
> I can have as a standalone patch that can be applied to the code to enable support - if that helps

That would be better than nothing – thank you.

AlmaLinux/RHEL/Rocky Linux 8 reach EOL on May 31, 2029.

eakraly:
Committed as a patch in #1952

PR #1918, opened 2026-05-18
Update docs for drop-invalid-packets and response-origin-only-with-rfc5780

### LINKED ISSUE #1917: Documentation update re-introduced deprecated --response-origin-only-with-rfc5780
`--response-origin-only-with-rfc5780` option still exists but doesn't do anything https://github.com/coturn/coturn/blob/7cde430a988feaf8d5f7b3cde455d696981f88a4/src/apps/relay/mainrelay.c#L2480
* Deprecated and removed from (some, but not all) documentation in https://github.com/coturn/coturn/pull/1690
* Added back to documentation in https://github.com/coturn/coturn/pull/1903, should instead be cleaned up

csiltala:
I attempted to use `make-man.sh` and it added a ton of man macro formatting back in,
since it was a minor targeted fix I just updated `turnserver.1` manually.

Fixes: https://github.com/coturn/coturn/issues/1917


eakraly:
Thank you @csiltala

PR #1921, opened 2026-05-22
Pr1588

eakraly:
This PR builds on top of #1588 and reworks the 401 rate-limiting feature, addressing correctness/security issues, improving tests, and switching to a lock-free design.

## What it does

Adds opt-in, per-source-IP rate-limiting of **401 Unauthorized** responses on UDP. This mitigates reflection/amplification abuse, where an attacker spoofs a victim's source address to bounce 401 challenges off the server. When a source exceeds its budget within the window, further 401 responses to that source are suppressed.

## CLI flags

- `--401-ratelimit` — enable the feature (off by default; opt in explicitly).
- `--401-req-limit=<count>` — max 401 responses per source IP per window (default `1000`).
- `--401-window=<seconds>` — window length in seconds (default `120`).

All three are documented in `README.turnserver` and the `turnserver` usage text.

## Design — lock-free rate limiter (`src/server/ns_turn_ratelimit.{c,h}`)

- Fixed-size, statically-allocated direct-mapped table (`4096` buckets, ~48 KiB resident) — no malloc/free, no list scan, no global mutex on the hot path.
- Each bucket holds atomics: a 32-bit tag (hash of the source IP with the **port stripped**, so attackers can't evade by rotating source ports), the window start, a counter, and a one-shot `logged` flag.
- The address is never stored as a pointer — only as a hash tag — so there is no use-after-free surface.
- On hash collision the bucket rotates to the new tag; the previous occupant simply gets a fresh window next time. Under attack, collisions converge to "drop", which is the desired behavior.
- Trades precision for a fully lock-free path: this is acceptable because the goal is abuse mitigation, not exact accounting.

## Integration

- `handle_turn_command` consumes a token only on requests that actually produce a 401, and only for `UDP_SOCKET` clients (reflection requires a spoofable source).
- The "401 rate-limit exceeded" log line is self-rate-limited to **one message per (bucket, window)** via the `first_drop` flag, so a flood does not cause log-write amplification.
- The three params are stored as **pointers** into `turn_params` and dereferenced at use sites, so the CLI flags affect every relay thread without per-thread copies.

## Correctness/security fixes (vs. the original #1588)

- The legacy struct typed `ratelimit_401_requests` as `bool` instead of `bool *`, which silently truncated the parameter pointer in `init_turn_server()` and left the feature effectively always-on. All three fields are now consistently pointers.
- Invalid `--401-req-limit`/`--401-window` values (≤ 0) are rejected with a warning and fall back to defaults.

## Tests

- New driver `examples/run_tests_ratelimit_401.sh` exercises the flags end-to-end:
  - **positive:** `--401-req-limit=1` drives a bounded `turnutils_uclient` session that crosses the threshold; the server must emit exactly one "401 rate-limit exceeded" log line.
  - **negative:** `--401-req-limit=100000` with the same traffic must not emit the log line.
- Split out of `run_tests.sh` so the rate-limit server fixture is fully isolated from the protocol-test fixture (a regression in either suite can't mask the other). Skipped on macOS, where loopback UDP relay flakiness makes log-line accounting unreliable; Linux CI is the canonical target.

## Build wiring

`ns_turn_ratelimit.{c,h}` added to both the CMake (`src/server/CMakeLists.txt`) and legacy Makefile (`Makefile.in`) source lists.


julialongtin:
Hey, just checking why this was closed; looked like good work?

PR #1948, opened 2026-06-18
validate hmackey length in sqlite_get_user_key before hex decode

alhudz:
`Repro:` store a `turnusers_lt` row whose `hmackey` is shorter than the key size (e.g. `00112233`) and authenticate as that user; `test_short_hmackey_rejected` in `tests/test_sqlite_dbd.c` reproduces it.
`Cause:` `sqlite_get_user_key` hands the column straight to `convert_string_key_to_binary`, which reads `sz * 2` hex chars unconditionally (`sz == 16` for `SHATYPE_DEFAULT`), so a short value over-reads past the column buffer and a NULL column dereferences. The `mysql`/`pgsql`/`redis`/`mongo` drivers already length-check before decoding; the `sqlite` one didn't.
`Fix:` reject NULL and under-length keys before decoding, matching the `pgsql` driver.


eakraly:
Thank you @alhudz !

PR #1950, opened 2026-06-20
fix signed-char index out-of-bounds read in base64_decode

alhudz:
`Repro:` with `--mobility` enabled, send a TURN Refresh carrying a `MOBILITY-TICKET` whose value is a 4-byte string with the high bit set (e.g. `\x80\x80\x80\x80`); the relay decodes it through `string_to_mobile_id` -> `base64_decode`.
`Cause:` `base64_decode` indexes `decoding_table[(int)data[i]]`, so a byte >= `0x80` sign-extends to a negative index and reads up to 128 bytes before the 256-byte table. `build_base64_decoding_table` already casts to `unsigned char` when filling it. Empty input also probes `data[input_length-1]`/`[-2]` before the buffer.
`Fix:` cast to `unsigned char` before the table lookup and reject empty input.

Added `tests/test_base64.c`; the high-bit case trips ASan (`heap-buffer-overflow` read in `base64_decode`) on the current tree and passes with the change.

### DIFF
--- a/src/apps/common/apputils.c
+++ b/src/apps/common/apputils.c
@@ -1360,6 +1360,11 @@ unsigned char *base64_decode(const char *data, size_t input_length, size_t *outp
     return NULL;
   }
 
+  if (input_length == 0) {
+    *output_length = 0;
+    return NULL;
+  }
+
   *output_length = input_length / 4 * 3;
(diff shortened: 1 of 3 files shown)


eakraly:
Thank you @alhudz

PR #1979, opened 2026-07-10
Wpb 26951 federation coturn preserve 9191

siberijah:
Ticket: https://wearezeta.atlassian.net/browse/WPB-26951

## What’s happening
Coturn servers communicate over UDP port **9191**. Kubernetes node NAT rewrites the source port with --random-port on AWS VPC CNI, so replies no longer match the connection. The secure handshake fails, causing **one-way audio**.

## Fix
Add an SNAT rule that preserves source port **9191**.

## Verify the rule
Run on a coturn node:
```bash
iptables-nft -t nat -S POSTROUTING | grep 9191
```
Expected:
```bash
-A POSTROUTING -p udp --sport 9191 -j SNAT --to-source 10.20.22.106:9191
```
The rule must appear **before** `AWS-SNAT-CHAIN-0`.

## Verify during a test call
```bash
conntrack -L -p udp --orig-port-src 9191
```
Good:
```text
sport=9191 dport=9191 ... sport=9191 dport=9191
```
Bad: any random rewritten port, such as `sport=54123`.
The actual pass condition is **audio working in both directions**.

## Caveat

The fix must exist on both coturn nodes handling the call


eakraly:
oopsie

PR #2032, opened 2026-08-03
Keep the legacy random nonce at 16 chars

eakraly:
## What

The legacy (non-`--stateless-nonce`) challenge nonce has been 24 hex chars instead of 16 since #2002, on every 64-bit server — including those that never enabled `--stateless-nonce`.

## Why

`generate_random_challenge_nonce` writes each `turn_random_number()` draw with `snprintf` and relies on the size bound to clip it. `"%08lx"` of a 64-bit draw is *up to 16* hex chars, and `"%04x"` of a 32-bit draw is 8 — the bound is what cuts each write back to 8. That bound was `NONCE_MAX_SIZE`, which [grew from 17 to 25](https://github.com/coturn/coturn/blob/master/src/server/ns_turn_session.h) when it was redefined as `TURN_STATELESS_NONCE_SIZE` to hold the 24-char stateless nonce. The extra digits then went through:

| Platform | `sizeof(long)` | Branch | Before #2002 | On master |
|---|---|---|---|---|
| Linux/macOS x86-64, arm64 | 8 | `%08lx` | 16 chars | **24 chars** |
| Windows x64, 32-bit | 4 | `%04x` | 16 chars | **20 chars** |

Reproduced standalone with the exact expressions from the two loops:

```
sizeof(long)=8
old NONCE_MAX_SIZE=17: len=16 'deadbeefdeadbeef'
new NONCE_MAX_SIZE=25: len=24 'deadbeefdeadbeefcafebabf'
```

## Impact

Low, and not a security issue:

- **No overflow** — `ss->nonce` is `NONCE_MAX_SIZE` (25) bytes; the maximum write is 24 + NUL.
- **No auth breakage** — validation is a `strcmp` against `ss->nonce`, so the longer nonce round-trips.
- **Entropy went up, not down** — 96 bits vs. the previous 64 on LP64.
- **Still RFC-legal** — RFC 8489 §14.10 caps the nonce below 128 characters.

What it does break is a documented invariant: `ns_turn_session.h` and `create_challenge_response` both state that the two nonce formats *differ in length*, and `tests/test_stateless_nonce.c` asserts that "legacy random-style 16-char nonces are not stateless nonces". On 64-bit they no longer differ. Nothing branches on that length today (`turn_check_stateless_nonce` rejects on the MAC, not the length), but it is a trap for the next length-based fast path. Separately, an opt-in feature silently changed wire output for deployments that did not opt in.

## Change

- Add `TURN_RANDOM_NONCE_LENGTH` / `TURN_RANDOM_NONCE_SIZE` so each nonce format is sized by its own constant.
- Extract the generation into `generate_random_challenge_nonce()`, bounded by `TURN_RANDOM_NONCE_SIZE`, with a comment recording that the bound — not the format specifier — is what fixes the width.
- Cover both loops with a unit test in the existing server-core harness.

Behavior is restored byte-for-byte to pre-#2002: the truncation semantics are preserved rather than replaced with an explicit bit-selection, which would have picked different bits.

## Validation

- `ctest` — 18/18 pass on macOS/arm64, 17/17 in Linux/Docker.
- Reverting just the bound fails the new test with `Expected 16 Was 24`.
- `run_tests.sh`, `run_tests_conf.sh`, `run_tests_stateless_nonce.sh`, `run_tests_mobile.sh`, `run_tests_multiplex_peer.sh`, `run_tests_dtls_default.sh`, `run_tests_rfc5780.sh` — all clean on Linux; the first four also clean on macOS.
- `fuzzing/run-local.sh ASan 0 -runs=1` and `ASan 1 -runs=1` — clean.
- `clang-format-15` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/server/ns_turn_server.c
+++ b/src/server/ns_turn_server.c
@@ -3486,6 +3486,30 @@ static int need_stun_authentication(turn_turnserver *server, ts_ur_super_session
   return 0;
 }
 
+/* Fill `nonce` (at least NONCE_MAX_SIZE bytes) with the legacy random
+ * challenge nonce: TURN_RANDOM_NONCE_LENGTH lowercase hex chars.
+ *
+ * Each snprintf is bounded by what is left of TURN_RANDOM_NONCE_SIZE, not of
+ * the destination buffer: `%08lx` of a 64-bit draw is up to 16 chars and
+ * `%04x` of a 32-bit draw is 8, so it is the bound that clips every write back
+ * to the 16-char format. Bounding by the (larger) stateless-nonce buffer would
+ * let those digits through and emit a 24-char nonce instead. */
+static void generate_random_challenge_nonce(uint8_t *nonce) {
+  if (TURN_RANDOM_SIZE == 8) {
+    for (int i = 0; i < (NONCE_LENGTH_32BITS >> 1); i++) {
+      uint8_t *s = nonce + 8 * i;
+      const uint64_t rand = (uint64_t)turn_random_number();
+      snprintf((char *)s, TURN_RANDOM_NONCE_SIZE - 8 * i, "%08lx", (unsigned long)rand);
+    }
+  } else {
+    for (int i = 0; i < NONCE_LENGTH_32BITS; i++) {
+      uint8_t *s = nonce + 4 * i;
+      const uint32_t rand = (uint32_t)turn_random_number();
+      snprintf((char *)s, TURN_RANDOM_NONCE_SIZE - 4 * i, "%04x", (unsigned int)rand);
+    }
+  }
+}
+
 static int create_challenge_response(ts_ur_super_session *ss, stun_tid *tid, int *resp_constructed, int *err_code,
                                      const uint8_t **reason, ioa_network_buffer_handle nbh, uint16_t method) {
   size_t len = ioa_network_buffer_get_size(nbh);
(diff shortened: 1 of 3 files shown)


eakraly:
Strengthened the unit test — as originally written it could not have caught a change to the constant itself.

The assertion read `TEST_ASSERT_EQUAL_size_t(TURN_RANDOM_NONCE_LENGTH, strlen(nonce))`, using the same macro the generator bounds itself with. The two move in lockstep, so any redefinition of `TURN_RANDOM_NONCE_LENGTH` (or of `NONCE_LENGTH_32BITS` beneath it) changed the emitted nonce *and* the expected value together. Demonstrated by setting it to `NONCE_LENGTH_32BITS * 5`: the nonce became 20 chars and all 18 tests still passed.

The test now asserts against a literal `RANDOM_NONCE_WIRE_LENGTH 16` — 16 is the wire format, not whatever the generator's bound happens to be — plus a check that the macro agrees with it, and a check that nothing past the terminator was written (the generator gets a 25-byte `NONCE_MAX_SIZE` buffer but may only use its own 17).

Mutation-tested, each rebuilt from a forced clean object:

| Mutation | Result |
|---|---|
| clean tree | 18/18 pass |
| bound reverted to `NONCE_MAX_SIZE` (the original bug) | FAIL — `Expected 16 Was 24` |
| `TURN_RANDOM_NONCE_LENGTH` redefined to 20 | FAIL — `Expected 16 Was 20` (previously passed) |
| generator scribbles at `nonce[20]` | FAIL — `Expected 0xFF Was 0x41` |

Production code is unchanged from the first push. `ctest` 18/18 on macOS/arm64 and 17/17 in Linux/Docker; `clang-format-15` clean.

PR #2057, opened 2026-08-22
Build docker images with optimizations on

eakraly:
Enable `-O3` for docker images
Enable optimizations where makes sense in cmake builds


fippo:
that is cheap 5% performance from a benchmark ;-)

Not that I would recommend coturn at CPU where this matters but ...

eakraly:
@fippo depends your specific setup but the biggest wins come from using --multiplex-peer where recvmmsg/sendmmsg/gso can work.
Without multiplexing peer - only receiving on UDP 3478 benefits from recvmmsg

PR #2060, opened 2026-08-24
fix: external ip must valid if the provider responses EDNS

tangxinfa:
Google may response IP and EDNS records, e.g.

    $ dig @ns1.google.com o-o.myaddr.l.google.com TXT -4 +short | tr -d \"
    edns0-client-subnet 103.252.16.0/24
    172.253.5.26

It pass the is_valid_ip() checking, but not a valid ip address, fix it by examine the first line only.


eakraly:
Hi @tangxinfa 
Thanks for the PR
It has a few issues:

1. In your example the wrong line will be selected
2. log redirect now misplaced
3. IPv6 branch in script also requires a fix

tangxinfa:
Minor fix: instead of blindly ignore extra lines of the command output, check if exactly one line.

Support IPv4 and IPv6.

PR #2067, opened 2026-08-30
Return allocation mismatch for ChannelBind without allocation

pannal:
When a client sends ChannelBind after its TURN allocation has expired or disappeared, the server should report Allocation Mismatch so the client can recreate the allocation. The current handler leaves the error unset and the generic response path emits Bad Request instead.

## What changes

- Return error 437, Allocation Mismatch, when ChannelBind has no valid allocation for its 5-tuple.
- Add a server-handler regression test for the expired-allocation case.

## Verification

The focused server test executable passes all 22 cases. The new regression test fails against the previous handler because it does not set the required 437 error.

### DIFF
--- a/tests/test_turn_server_send.c
+++ b/tests/test_turn_server_send.c
@@ -701,6 +701,25 @@ static void test_channel_bind_binds_the_peer_address(void) {
   TEST_ASSERT_TRUE(addr_eq(&peer, &(chn->peer_addr)));
 }
 
+/* RFC 8656 Section 7.3: requests other than Allocate require an existing
+ * allocation for the request's 5-tuple. A ChannelBind received after that
+ * allocation expires reports Allocation Mismatch rather than falling through
+ * to the generic Bad Request response. */
+static void test_channel_bind_without_a_valid_allocation_reports_mismatch(void) {
+  ioa_addr peer;
+  make_addr(&peer, PEER_A, PEER_PORT_A);
+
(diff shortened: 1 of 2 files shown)


tests/test_turn_server_send.c
```diff
@@ -701,6 +701,25 @@ static void test_channel_bind_binds_the_peer_address(void) {
   TEST_ASSERT_TRUE(addr_eq(&peer, &(chn->peer_addr)));
 }
 
+/* RFC 8656 Section 7.3: requests other than Allocate require an existing
+ * allocation for the request's 5-tuple. A ChannelBind received after that
+ * allocation expires reports Allocation Mismatch rather than falling through
+ * to the generic Bad Request response. */
```
eakraly: The rule this test pins lives in RFC 8656 **Section 5 ("General Behavior")**, not Section 7.3 (which is "Receiving an Allocate Success Response"):

> For all TURN messages (including ChannelData) EXCEPT an Allocate request, if the 5-tuple does not identify an existing allocation, then the message MUST either be rejected with a 437 Allocation Mismatch error (if it is a request) or be silently ignored (if it is an indication or a ChannelData message).

```suggestion
/* RFC 8656 Section 5: for any request other than Allocate, a 5-tuple that
 * does not identify an existing allocation is rejected with 437 (Allocation
 * Mismatch). A ChannelBind received after its allocation expires reports
 * that rather than falling through to the generic Bad Request response. */

PR #2073, opened 2026-09-01
hiredis: clean up failed async connection setup

junkilee80:
## Summary

- free non-NULL Hiredis async contexts returned with an error
- clean up partially initialized libevent state in attach and reconnect paths
- keep failed reconnect state retryable
- add real Hiredis/libevent fault-injection regression tests

## Problem

`redisAsyncConnect()` can return a non-NULL context with `ac->err` set. The
caller still owns that context, but the initial attach path did not free it,
while the reconnect path could treat it as a valid connection.

If either `event_new()` call failed, the initial attach path freed only its
wrapper, while reconnect retained a partial context/event state.

Cleanup must go through `redisAsyncFree()` because the registered libevent
cleanup callback owns the events. Freeing those events separately would result
in a double-free.

## Fix

- call `redisAsyncFree()` for Hiredis error contexts
- clean up the async context when either event allocation fails
- release attach-only wrapper strings after context cleanup
- reset the reconnect context to `NULL` so a later retry can succeed

## Testing

The new test uses the real Hiredis and libevent implementations. It wraps only
`event_new()` to inject deterministic read/write event allocation failures.

It covers:

- Hiredis error contexts during initial attach and reconnect
- read and write event allocation failures
- successful retry after reconnect setup fails

Results:

- same unpatched commit plus the regression test: 6/6 cases fail
- patched ASan/UBSan/LSan regression test: 6/6 cases pass
- patched ASan/UBSan/LSan CTest: 16/16 pass when excluding the pre-existing
  `test_turn_ports` teardown leak
- Release build and CTest: 17/17 pass
- `turnutils_rfc5769check`: all checks pass
- `git diff --check` and clang-format check pass

## Tool-assisted discovery

This was found while evaluating Vul-RAG and PailGen on coturn.

PailGen's initial event-failure cleanup draft conflicted with Hiredis' cleanup
callback and was rejected after ASan detected a double-free. The final cleanup
and regression tests were manually ownership-reviewed and validated using the
real libraries.

This is presented as a robustness and resource-cleanup fix. I am not claiming
a demonstrated remote security impact.

### DIFF
--- a/src/apps/relay/hiredis_libevent2.c
+++ b/src/apps/relay/hiredis_libevent2.c
@@ -260,6 +260,7 @@ redis_context_handle redisLibeventAttach(struct event_base *base, char *ip0, int
     return NULL;
   } else if (ac->err) {
     fprintf(stderr, "Error: %s:%s\n", ac->errstr, ac->c.errstr);
+    redisAsyncFree(ac);
     return NULL;
   }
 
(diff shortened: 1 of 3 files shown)


eakraly:
Thank you!

PR #2075, opened 2026-09-03
Silently ignore attributes after message-integrity

fippo:
as stipulated by
  https://datatracker.ietf.org/doc/html/rfc8489#section-9
which says
  Note that agents MUST ignore all attributes that follow MESSAGE-
  INTEGRITY, with the exception of the MESSAGE-INTEGRITY-SHA256 and
  FINGERPRINT attributes.

Claims this leads to auth bypass are overrated.


eakraly (approved):

eakraly:
Nice - thanks for the contribution @fippo !

